PR #10, opened 2022-07-13
Feature/ou tree clean

jordibc:
This PR addresses ticket https://app.clickup.com/t/314jnhm

It adds the commands `remove-orgunit-level` and `remove-orgunit-children`.

They both create a sql file that modifies a dhis2 database and removes all the organisation units below a given hierarchy level, or
all the children of a given organisation unit path.

The created materialized views help with gathering all the dependencies in a not-too-wild way.

The indices on the foreign key constraints with the slowest checks reduce much the time needed to run the generated sql file. The sql script should finish in a few minutes maximum (in my tests, below 3 minutes for the DEV database in worst-case scenarios).


src/scripts/cli.ts
```diff
@@ -3,13 +3,15 @@ import { run, subcommands } from "cmd-ts";
...
 import * as programs from "./commands/programs";
+import * as database from "./commands/database";
 
 export function runCli() {
     const cliSubcommands = subcommands({
         name: path.basename(__filename),
         cmds: {
             datasets: datasets.getCommand(),
             programs: programs.getCommand(),
+            database: database.getCommand(),
```
tokland, line 14: Can we use the model instead? `orgunits: `.

src/scripts/commands/database.ts
```diff
@@ -0,0 +1,205 @@
...
+                long: "output-file",
+                description: "Name of the output file that will contain the sql commands",
+            }),
+            overwrite: flag({
+                long: "overwrite",
+                description: "Overwrite ouput file if exists",
+            }),
+        },
+        handler: async args => {
+            const outputFile = args.outputFile ?? "remove_orgunits.sql";
```
tokland, line 29: [no strong opinion] Instead of hardcoded file paths, we can print to stdout.

src/scripts/commands/database.ts
```diff
@@ -0,0 +1,205 @@
...
+            const sqlCommands = removeOrgs(`path LIKE '${args.path}/%'`);
```
tokland, line 56: These 2 sub-commands are so similar that I wonder: does it make to create a single command with different options for each filter? example: `--from-level=LEVEL` / `--by-path=PATH`. Both filters could work together (as and AND, for example)

src/scripts/commands/database.ts
```diff
@@ -0,0 +1,205 @@
...
+        console.log(`Written: ${outputFile}`);
+    } else {
+        console.error(`ERROR. Output file already exists: ${outputFile}`);
+        console.error("Use --overwrite if you want to overwrite it anyway.");
+    }
+}
+
+// Return an SQL string that would remove all the organisation units that
+// satisfy the given whereCondition.
+function removeOrgs(whereCondition: string): string {
```
tokland, line 84: Following clean architecture guidelines, from `safeWrite` to the end should not live here. I'd create (assuming we only have one command with options):

- Use case: src/domain/usecases/CreateSQLForOrgUnitRemovalUseCase.ts 
- Repositories: interface -> src/domain/repositories/OrgUnitsRepository.ts (getDeleteScript) + src/data/OrgUnitD2Repository.ts (with the implementation)

src/scripts/commands/database.ts
```diff
@@ -0,0 +1,205 @@
...
+    return `
```
tokland, line 85: No strong opinion, but I am a bit wary of big chunks of interpolated strings. If you prefer to keep it like this, no problem. The alternative is to create a separate: `somefile.template.sql`, read the contents, and interpolate some namespace (for example: using _.template)

src/scripts/commands/database.ts
```diff
@@ -0,0 +1,205 @@
...
+
+
+\\echo -- Deleting orgunits themselves...
+
+DELETE FROM organisationunit WHERE organisationunitid IN (SELECT * FROM orgs);
+
+
+\\echo -- Cleaning up...
+
+DROP MATERIALIZED VIEW orgs CASCADE;`;
```
tokland, line 202: EOF missing

src/scripts/commands/database.ts
```diff
@@ -0,0 +1,205 @@
...
+import fs from "fs";
+import { command, subcommands, positional, flag, option, optional, string } from "cmd-ts";
+import { HierarchyLevel, OrgUnitPath } from "scripts/common";
+
+export function getCommand() {
+    const removeOrgUnitLevelCommand = command({
+        name: "remove-orgunit-level",
+        description: "Create a sql file that removes organisation units below a hierarchy level",
+        args: {
+            level: positional({
```
tokland, line 14: Usually, named options are more clear, without the `optional` , they will be compulsory.

src/scripts/commands/database.ts
```diff
@@ -0,0 +1,205 @@
...
+            safeWrite(outputFile, sqlCommands, args.overwrite);
+            process.exit(0);
+        },
+    });
+
+    const removeOrgUnitChildrenCommand = command({
+        name: "remove-orgunit-children",
+        description: "Create a sql file that removes all children of an organisation unit",
+        args: {
+            path: positional({
```
tokland, line 40: Same, here it's even less clear what the positional argument means, so better to have it named.

src/scripts/commands/database.ts
```diff
@@ -0,0 +1,205 @@
...
+
+DELETE FROM organisationunit WHERE organisationunitid IN (SELECT * FROM orgs);
+
+
+\\echo -- Cleaning up...
+
+DROP MATERIALIZED VIEW orgs CASCADE;`;
+}
+// Maybe this is worth checking too:
+// https://github.com/dhis2/dhis2-utils/blob/master/resources/sql/delete_orgunittree_with_data.sql
```
tokland, line 205: What's the target DHIS2 version? Tried the script, it didn't find some of the tables (2.36.8):

```
$ node dist/index.js database remove-orgunit-level 3
```

Output: http://download.zaudera.com/public/remove_orgunits.sql.out

tokland (commented):
That's a cool functionality to have, it will come in handy in many Samaritans projects. Some comments:

src/scripts/cli.ts
jordibc: Sure. This is what I had in mind though:

The way I imagined it, the file ("model") was related to `// Commands related to dhis2 database direct manipulation.` (From the old first line in database.ts) So I imagined it would extend in the future towards more subcommands related to database manipulation via SQL strings, not necessarily related to orgunits.

But I obviously trust your intuition more on this matters, so I'll change it.

src/scripts/commands/database.ts
jordibc: Yes, I like it better too. Done in 52f4ab802ac41ecf2049ae992b70785aa1062106 (with other things).

src/scripts/commands/database.ts
jordibc: Yes, makes sense. This came from not being that familiar with cmd-ts. Done as part of 52f4ab802ac41ecf2049ae992b70785aa1062106

src/scripts/commands/database.ts
jordibc: Yes, I like it that way too, I just didn't know about _.template, thanks! Not sure about its proper location, though. I added it to `./src/data/` in 5ced1758166c019f10c8572e861edeb175f2728c .

src/scripts/commands/database.ts
jordibc: Fixed 5ced1758166c019f10c8572e861edeb175f2728c

src/scripts/commands/database.ts
jordibc: I changed it so you can add either option, and it will fail if both are missing. I suspect there may be a more idiomatic way than the way I did it, though.

src/scripts/commands/database.ts
jordibc: Actually, it worked :) The output shows it selected 0 org units, and deleted 0 org units. The only errors are related to some tables not existing, which is okay, and you can see a previous output text warning that it may happen and it is still okay (that is related to other tables existing in other dhis2 versions).

I have tested it with several DBs with 2.36 and 2.38.

src/scripts/commands/database.ts
jordibc: I've been trying for quite a while to wrap my head around this one. I don't get it. This is the only thing I have not addressed. I think I would need help with this.

tokland: You can ask Nacho how to tackle this, maybe Jorge can step in. The code is very located and won't hurt a non-clean-arch approach for now, but in the long run, we should keep all the package on clean-arch.

src/scripts/commands/database.ts
tokland: I see, although the org unit is deleted from the DB, the API (this Maintenance) still show it. I guess that's normal, because of the cache and we need to restart the instance.

src/scripts/commands/database.ts
tokland: It looks good to me

src/scripts/commands/database.ts
ifoche: @jordibc @tokland I think Arnau's idea is very good. Jordi, could you ask Jorge to check this PR and contribute with the changes that would make this clean? If you could afterwards have a short meeting with him so he can explain it to you, that would be great. Thank you!

xurxodev: Hi @jordibc, @ifoche asked me to review this PR about clean architecture.

I agree with @tokland, the code in the handle should be moved to a use case.

The goal of clean architecture is separate responsibilities and hide details of implementation. 

We are going to review problems with the current implementations.

* The logic under org units command is coupled to cmd-ts
* The infrastructure logic to read and write files is coupled to fs
* Here there are three responsibilities:
   * Presentation -> cmd-ts related code
   * Domain -> The coordination of the actions to realize:
       1. validations
       2. Create where condition
       3. get the template
       4. Create sqlCommands 
       5. Save sqlCommands
       6. Return sqlCommands
   * Infrastructure o data -> read the template from disk and save the sqlCommands to disk

But all three responsibilities are in the same file in a coupled approach, consequences:

* Homogeneity -> other commands in this project follow clean architecture less this. This creates a problem, for example, the next person will have doubts about when to use cases o not.
* You are the code creator but for example, if another person that doesn't know the code modifies it, how are the 3 responsibilities together maybe wanting to change one creates a bug in another. The classic problem of fixing one thing and breaking another.
* You can not create an integration test that executes the logic in the org units command without using cmd-ts and verifying the final output is the expected
* You can not create unit tests that verify that expected validations are realized

How to migrate to clean architecture:
- Presentation: read args, invoke to the use case, the exit of the process when the use case returns errors o throws exceptions, the exit of the process when to the finish
- Domain: create a use case with the coordination tasks
- Infrastructure or data: create a repository to read and write files on disk

I think that reviewing code in other commands in this same project can help.

tokland (approved):
Actions:

- [x] Merged with development
- [x] Added some extras dependencies that I detected when testing the script against docker.eyeseetea.com/samaritans/dhis2-data:2.36.10.1-sp-cpr
- [x] Apply clean architecture
- [x] Some refactors

PR #11, opened 2022-07-14
Run program rules (fixes+refactors)

tokland:
Closes https://app.clickup.com/t/28ec4rq

This PR will contain all the modifications required by the feedback we get from the client: https://app.clickup.com/t/2ad8fu9


tokland:
- [ ] Confirmation by the client that it's working in their scenario

PR #15, opened 2022-10-10
Feature/Script to copy org units from 1 to n datasets

nshandra:
### :pushpin: References

Issue: [Script to copy org units from 1 to n datasets](https://app.clickup.com/t/3jn9vg7)

### :memo: Implementation

Command:
```console
shell:~$ yarn start datasets copy-org-units \
    --url='http://USER:PASSWORD@localhost:8080' \
    --origin-dataset=DiMntK7qKZQ \
    --destination-datasets=Z3tlf5sqWiK,jHF49Vvup66 \
    [--replace, -r]

shell:~$ yarn start datasets copy-org-units \
    --url='http://USER:PASSWORD@localhost:8080' \
    -o=DiMntK7qKZQ \
    -d=Z3tlf5sqWiK,jHF49Vvup66 \
    [--replace, -r]
```

Notes:
- `origin-dataset` and `destination-datasets` has a length check to match DHIS2 ID length. 
- If a destination data set contains all the origin OUs no action is taken and a warning is shown.
- The `--replace` flag allows to overwrite the destination Organisation Units. If `--replace` is set, and the OUs are **identical** in the origin and destination data sets, no action is taken and a warning is shown.
- If no data set is modified a warning is shown.
- The "simple-node-logger" lib for `utils/log.ts` gets replaced by an implementation by @tokland.


README.md
```diff
@@ -94,3 +94,26 @@ shell:~$ yarn start datasets compare \
...
+    --url='http://USER:PASSWORD@localhost:8080' \
+    --origin-dataset=DiMntK7qKZQ \
+    --destination-datasets=Z3tlf5sqWiK,jHF49Vvup66 \
+    [--replace, -r]
+
+shell:~$ yarn start datasets copy-org-units \
+    --url='http://USER:PASSWORD@localhost:8080' \
+    -o=DiMntK7qKZQ \
+    -d=Z3tlf5sqWiK,jHF49Vvup66 \
+    [--replace, -r]
```
tokland, line 111: [minor] [subjective] I wouldn't duplicate the example showing the short options, the user can check this with `--help`.

src/data/DataSetsD2Repository.ts
```diff
@@ -28,6 +28,16 @@ export class DataSetsD2Repository implements DataSetsRepository {
         }
     }
 
+    async post(data: object, options: object): Promise<MetadataResponse> {
+        const response = await this.api.metadata.post(data, options).getData();
+
+        if (response.status !== "OK") {
+            console.error(JSON.stringify(response.typeReports, null, 4));
+        }
```
tokland, line 36: We have `runMetadata` (https://github.com/EyeSeeTea/d2-tools/blob/feature/datasets_copy-org-units/src/data/dhis2-utils.ts#L21). Note that, in case of an error, this function performs a rejection.

src/domain/usecases/CopyDataSetsOUUserCase.ts
```diff
@@ -0,0 +1,96 @@
+import _ from "lodash";
+import { DataSetsRepository } from "domain/repositories/DataSetsRepository";
+import { PostOptions, Ref } from "types/d2-api";
+import { DataSet } from "domain/entities/DataSet";
+import { OUCopyResult } from "domain/entities/OUCopyResult";
+
+export class CopyDataSetsOUUserCase {
+    constructor(
+        private dataSetsRepository: DataSetsRepository
+    ) { }
```
tokland, line 10: I see `prettier` is installed, but there is no `prettify` command, added. Now `yarn prettify` applies the code format. 

It's usually very handy to configure vscode to automatically apply format on save. Setting: "Format on Save"

src/domain/usecases/CopyDataSetsOUUserCase.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    async execute(options: {
+        originDataset: string;
+        destinationDatasets: string[];
+        replace: boolean | undefined;
+    }): Promise<OUCopyResult> {
+        const { originDataset, destinationDatasets, replace } = options;
+
+        console.debug(`Replace the destination OU: ${replace ? "True" : "False"}`);
```
tokland, line 19: We can use `log.debug` (from utils/log)

src/domain/usecases/CopyDataSetsOUUserCase.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        const datasets = await this.dataSetsRepository.get([originDataset, ...destinationDatasets]);
+
+        const origDataset = datasets[originDataset];
+        const destDatasets = destinationDatasets.map((id) => datasets[id]);
+
+        if (!origDataset || !destDatasets) throw new Error("Missing DataSets");
```
tokland, line 26: Empty JS arrays are always truthy (unlike Python), so this condition will be never met. You can use `_.compact` and check that the resulting array has the expected length, for example. `if (!destDataSet` will be no longer necessary.

src/domain/usecases/CopyDataSetsOUUserCase.ts
```diff
@@ -0,0 +1,96 @@
...
+            if (!dataSetsEqual) {
+                if (replace) {
+                    data.push({ ...destDataSet, organisationUnits: origDataset.organisationUnits })
+                } else {
+                    data.push(mergeDataSetOUs(origDataset, destDataSet));
+                }
+            } else {
+                console.debug(`DataSet with ID:${destDataSet.id} already contains all the OUs.`);
+            }
+        });
```
tokland, line 43: Functional programming. Pattern: empty array + forEach + push -> `_(inputArray).map(.... returns some value or undefined ...).compact().value()`

src/domain/usecases/CopyDataSetsOUUserCase.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        let result: OUCopyResult;
+        if (!_.isEmpty(data)) {
+            const metadata = { dataSets: [...data] };
```
tokland, line 47: Why the copy? `[...data]`. It should not be touched by the post.

src/domain/usecases/CopyDataSetsOUUserCase.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (replace) {
+        differences = _.isEqual(OUs1ToDiff, OUs2ToDiff);
+    } else {
+        const filtered = OUs2ToDiff.filter((ou) => {
+            return ou.id === OUs1ToDiff.find(ou2 => ou2.id === ou.id)?.id;
+        })
+        differences = _.isEqual(OUs1ToDiff, filtered)
+    }
+
+    return differences;
```
tokland, line 75: [minor] It's ok as is. Just as a note, it's also idiomatic that, being the last expression of a function, we have a return inside the branches (it's not considered a "multiple exit points" red flag)

src/domain/usecases/CopyDataSetsOUUserCase.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+function getDataSetWithSortedOUs(dataSet: DataSet): DataSet["organisationUnits"] {
+    return sortById(dataSet.organisationUnits);
+}
+
+function mergeDataSetOUs(orgDataset: DataSet, destDataSet: DataSet): DataSet {
+    const mergedOUs: DataSet["organisationUnits"] = _.uniqBy(
+        [...orgDataset.organisationUnits, ...destDataSet.organisationUnits],
+        "id"
```
tokland, line 85: Unfortunately, this style of property is not type-safe in lodash. So we typically write `obj => obj.id` (some projects have Base.ts -> getId to have this ready to use).

src/scripts/common.ts
```diff
@@ -34,3 +34,20 @@ export const StringsSeparatedByCommas: Type<string, string[]> = {
...
+
+export const IDString: Type<string, string> = {
+    async from(str) {
+        if (_(str).isEmpty()) throw new Error("Value cannot be empty");
+        if (str.length !== 11) throw new Error("ID must be 11 char long");
+        return str;
+    },
+};
+
+export const IdsSeparatedByCommas: Type<string, string[]> = {
```
tokland, line 46: Better to use the existing function`src/scripts/common.ts`. The string validation can be added there.

src/domain/repositories/DataSetsRepository.ts
```diff
@@ -1,7 +1,8 @@
 import { DataSet } from "domain/entities/DataSet";
-import { Id } from "types/d2-api";
+import { Id, MetadataResponse } from "types/d2-api";
 
 export interface DataSetsRepository {
     get(ids: Id[]): Promise<Record<Id, DataSet>>;
+    post(data: object, options: object): Promise<MetadataResponse>;
```
tokland, line 6: `object` is too generic (only to use when you don't interpret the entity, but here we have a `DataSet`). 

The same applies to `options`, we should specific supported options. Typically, we don't expose all the d2-api options, only the ones we need.

tokland, line 6: `MetadataResponse` is not a domain entity, it's an implementation detail of d2-api. Returns the most simple value needed by your usecase.

src/domain/usecases/CopyDataSetsOUUserCase.ts
```diff
@@ -0,0 +1,96 @@
...
+    constructor(
+        private dataSetsRepository: DataSetsRepository
+    ) { }
+
+    async execute(options: {
+        originDataset: string;
+        destinationDatasets: string[];
+        replace: boolean | undefined;
+    }): Promise<OUCopyResult> {
+        const { originDataset, destinationDatasets, replace } = options;
```
tokland, line 17: It's idiomatic to set the default values here: `...., replace = false } = options`

src/domain/usecases/CopyDataSetsOUUserCase.ts
```diff
@@ -0,0 +1,96 @@
...
+import { DataSet } from "domain/entities/DataSet";
+import { OUCopyResult } from "domain/entities/OUCopyResult";
+
+export class CopyDataSetsOUUserCase {
+    constructor(
+        private dataSetsRepository: DataSetsRepository
+    ) { }
+
+    async execute(options: {
+        originDataset: string;
```
tokland, line 13: Minor: `string` -> `Id`

tokland (commented):
The code looks good! And the feature worked in my tests. Just some notes about prettification, idiomatic JS, and clean arch.

README.md
nshandra: Yeah, i was unsure about that. Better to add a text note about the short options.

src/domain/usecases/CopyDataSetsOUUserCase.ts
nshandra: I have Format on Save activated, just needed to switch to Prettier formatter.

src/domain/usecases/CopyDataSetsOUUserCase.ts
nshandra: On further testing, there's no need to use uniqBy instead of uniq.

src/domain/usecases/CopyDataSetsOUUserCase.ts
nshandra: Indeed, no need for the shallow copy.

src/domain/usecases/CopyDataSetsOUUserCase.ts
nshandra: That's what I tried at first, but it didn't showed up in the terminal. Turns out that log doesn't work (except log.info) without a .env file. That's how its defined in utils/log.ts so i guess its intended this way.

Switching console.* to log.*

src/domain/usecases/CopyDataSetsOUUserCase.ts
nshandra: That if its not for checking if they are empty, but to deal with this:
TS2345: Argument of type 'DataSet | undefined' is not assignable to parameter of type 'DataSet'.
  Type 'undefined' is not assignable to type 'DataSet'.
datasets[originDataset] always will exists because this.dataSetsRepository.get will fail previously otherwise.

src/domain/usecases/CopyDataSetsOUUserCase.ts
nshandra: Done

src/domain/usecases/CopyDataSetsOUUserCase.ts
nshandra: Done.

src/scripts/common.ts
nshandra: You mean to use StringsSeparatedByCommas and add a handler in the userCase?
IdsSeparatedByCommas is a bit redundant, but that way we get a consistent options error messaging.
```
  datasets copy-org-units --url=http://admin:district@localhost:8080 --origin-dataset=LthvL9t2acW --destination-datasets=RvS8hay27O2,RvS8hay27O3,RvS8hay27O -r
                                                                                                  ^ IDs must be 11 char long
```

src/data/DataSetsD2Repository.ts
nshandra: Switched to runMetadata. Added try/catch.

src/domain/repositories/DataSetsRepository.ts
nshandra: Switched to: `post(data: DataSetMetadata): Promise<OUCopyResult>;`.

nshandra: Changed it to return the post status.

src/domain/usecases/CopyDataSetsOUUserCase.ts
tokland: Yeah, `info` is the default log level.

https://github.com/EyeSeeTea/d2-tools/blob/feature/datasets_copy-org-units/src/utils/log.ts#L12

It can be changed using an environment variable: `LOG_LEVEL=debug yarn start ...`

I'll add a note in the README

src/domain/entities/DataSetMetadata.ts
```diff
@@ -0,0 +1,5 @@
+import { DataSet } from "./DataSet";
+
+export interface DataSetMetadata {
```
tokland, line 3: [subjective] Typically, we group very related interfaces in the "parent" entity (`DataSet`), but it's ok like this.

src/domain/repositories/DataSetsRepository.ts
```diff
@@ -1,8 +1,10 @@
 import { DataSet } from "domain/entities/DataSet";
-import { Id, MetadataResponse } from "types/d2-api";
+import { DataSetMetadata } from "domain/entities/DataSetMetadata";
+import { OUCopyResult } from "domain/entities/OUCopyResult";
```
tokland, line 3: [minor] As before, having too many small entities clutters the folder. Types used only in the use case, can be exported in the use case itself.

src/domain/usecases/CopyDataSetsOUUserCase.ts
tokland: Sorry, it was confusing, I was referring to connected, but different things. I meant that if you use a compact on the array, then you remove undefined, and later the check is not necessary. But the way you've refactored it (keep T | undefined and check inside the map) is perfectly fine.

src/domain/usecases/CopyDataSetsOUUserCase.ts
```diff
@@ -61,29 +62,26 @@ export class CopyDataSetsOUUserCase {
...
 
 function mergeDataSetOUs(orgDataset: DataSet, destDataSet: DataSet): DataSet {
-    const mergedOUs: DataSet["organisationUnits"] = _.uniqBy(
-        [...orgDataset.organisationUnits, ...destDataSet.organisationUnits],
-        "id"
-    );
+    const mergedOUs: DataSet["organisationUnits"] = _.uniq([
+        ...orgDataset.organisationUnits,
+        ...destDataSet.organisationUnits,
+    ]);
```
tokland, line 84: `_.uniqBy` was ok (objects built separately are never equal in JS), I meant: `_.uniqBy(xs, x => x.id)`

src/domain/entities/DataSetMetadata.ts
nshandra: Ok, wasn't sure if it was better to keep them separated. Moved to entities/DataSet

src/domain/repositories/DataSetsRepository.ts
nshandra: Moved OUCopyResult type to DataSetsRepository.ts

src/domain/usecases/CopyDataSetsOUUserCase.ts
nshandra: Ok, changed it to _.uniqBy. But in this case _.uniq compares the object values (uniq -> SameValueZero -> SameValueNonNumber), so, the id value.

src/domain/usecases/CopyDataSetsOUUserCase.ts
nshandra: Used `_.compact` to lessen the check clutter.

tokland (approved):
All good! The feature works for me, code :+1:

nshandra:
Thanks @tokland

PR #20, opened 2023-01-30
Feature/add organisation unit groups

idelcano:
https://app.clickup.com/t/2tdt5x1

This issue ended a long time ago but it was not PRed by mistake waiting for some performance tests with events that are independent of the issue so I link it and leave the PR created.

It consists of adding the possibility of using orgunitgroups in addition to org units to the program rules script.


tokland (approved):
Feature worked well, I just did some code refactors (Ina, if you have any doubt, let me know)

PR #22, opened 2023-03-10
Feature/send email to users

deeonwuli:
### :pushpin: References

-   **Issue:** [Create script in d2-tools to send an email to a list of users](https://app.clickup.com/t/865bvymky)

### :memo: Implementation
- Command: `send-user-info-notification` - Send an email read from a JSON file to a list of users in a CSV file:


src/domain/repositories/RecipientRepository.ts
```diff
@@ -1,7 +1,8 @@
 import { Async } from "domain/entities/Async";
 import { Id } from "domain/entities/Base";
-import { Email } from "domain/entities/Notification";
+import { Email, UsernameEmail } from "domain/entities/Notification";
 
 export interface RecipientRepository {
     getByIds(ids: Id[]): Async<Email[]>;
+    getByUsernames(usernames: string[]): Async<UsernameEmail[]>;
```
tokland, line 7: Typically, both `get` would return the same type, here we have `Email` and the new `UsernameEmail`. I think it's a good to have `UsernameEmail`, but with some refactors. I propose this general change:

- [ ] Rename RecipientRepository (and the implementation) to UserRepository. 
- [ ] Create entity User.ts (with id, username, and email).

src/domain/repositories/UserInfoNotificationsRepository.ts
```diff
@@ -0,0 +1,6 @@
+import { Async } from "domain/entities/Async";
+import { UserInfoNotification } from "domain/entities/UserInfoNotification";
+
+export interface UserInfoNotificationsRepository {
+    get(userDetailsPath: string, emailContentJsonPath: string): Async<UserInfoNotification>;
```
tokland, line 5: The idea for the domain repository is to be as general as possible. We don't know where the data will come from, so it's better not to say "json" or "path". Proposal: 

`get(userDetailsId: string, emailContentsId: string): Async<UserInfoNotification>;`

And then in the data repository, the arguments can have the full name, and be interpreted as paths.

src/domain/usecases/SendUserInfoNotificationsUseCase.ts
```diff
@@ -0,0 +1,61 @@
...
+            .map(({ username }) => username)
+            .compact()
+            .value();
+        const recipients = await this.recipientRepository.getByUsernames(usernames);
+
+        log.debug(`Send user info email to: ${recipients.join(", ")}`);
+        const recipientsMap = recipients.reduce<{ [index: string]: UsernameEmail }>((acc, curr) => {
+            acc[curr.username] = curr;
+            return acc;
+        }, {});
```
tokland, line 45: A reduce can typically used with other lodash abstractions (groupBy, fromPairs, and so on). In this case, you are doing an indexing/keying, there are an abstraction for that -> `const recipientsMap = _.keyBy(recipients, recipient => recipient.username)`

src/domain/usecases/SendUserInfoNotificationsUseCase.ts
```diff
@@ -0,0 +1,61 @@
...
+
+        const userDetails: UserDetails[] = userInfoNotification.userInfos.map(userInfo =>
+            Object.assign(userInfo, recipientsMap[userInfo.username as keyof typeof recipientsMap])
```
tokland, line 48: `Object.assign` modifies the first object in place. It's better to create a new object: `{...obj1, ...obj2}`

tokland, line 48: This `as keyof typeof recipientsMap` means something is wrong. The access may not exist and return undefined. This case must be dealt with. A typical approach:

```
const userDetails = _(userInfoNotification.userInfos)
    .map((userInfo): UserDetails | undefined => {
        const user = recipientsMap[userInfo.username];
        return user ? { ...userInfo, email: user.email } : undefined;
    })
    .compact()
    .value();
```

src/domain/usecases/SendUserInfoNotificationsUseCase.ts
```diff
@@ -0,0 +1,61 @@
...
+        log.debug(`Get User Info and Email Content`);
+        const userInfoNotification = await this.userInfoNotificationRepository.get(
+            options.userDetailsFile,
+            options.emailContentFile
+        );
+
+        log.debug(`Get emails for each user in CSV file from DHIS2`);
+        const usernames = _(userInfoNotification.userInfos)
+            .map(({ username }) => username)
+            .compact()
```
tokland, line 37: Do we need this compact?

src/data/RecipientD2Repository.ts
```diff
@@ -33,4 +33,22 @@ export class RecipientD2Repository implements RecipientRepository {
...
+            .get({
+                users: {
+                    fields: { username: true, email: true },
+                    filter: { username: { in: usernames } },
+                },
+            })
+            .getData();
+
+        // @ts-ignore
+        return users;
```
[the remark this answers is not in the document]
deeonwuli: `user.username` isn't empty when i test with a 2.38 instance.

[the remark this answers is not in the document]
adrianq: @deeonwuli My bad, this should be tested with a 2.36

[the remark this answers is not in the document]
tokland: Oh, I see, in 2.38 it's inside userCredentials and at the root object.

src/domain/usecases/SendUserInfoNotificationsUseCase.ts
deeonwuli: it's not necessary.

PR #24, opened 2023-05-04
[NRC] Migrate dataset-configuration data set sections

tokland:
[DON'T MERGE] Close PR when task performed

Closes https://app.clickup.com/t/865c48cuk

- [x] Execute against PRO (by client)
- [x] Install app in PrO

Steps:

```
$ # Make a backup of database.
$ git clone https://github.com/EyeSeeTea/d2-tools
$ cd d2-tools
$ git checkout feature/nrc-ds-config-datasets-migration
$ nvm use v16.14.0
$ yarn install
$ npx ts-node src/scripts/nrc-ds-config-migrate-dataset-sections-codes.ts --url="https://USER:PASSWORD@gors.nrc.no"
```

Notes: 

- _USER_ must have permission to get and update data sets. 
-  Both _USER_ and _PASSWORD_ must be URL-encoded (for example, a `@` should be written as `%40`).  This online service can be used to transform the fields (note:  the separator `:` must be left as-is): https://www.freeformatter.com/url-encoder.html#before-output -> write the user or password and then click `[URL Encode]`.

Warnings:

- In some of the executions, I got "Batch update returned unexpected row count from update" error from the API. This happens when the section is being updated also in the UI. That's not a problem, just run the script again. The script can be run many times as desired, it will update only the required dataset sections. 
- The payload has reduced to the minimum, and the chunks reduce to 10 sections to avoid payload size problems (that we found in gorsdev for chunk=100. The max payload size depends not on DHIS2 but the web server -nginx, Apache- over it).

### DIFF
--- a/src/data/dhis2-utils.ts
+++ b/src/data/dhis2-utils.ts
@@ -53,10 +53,13 @@ export async function getInChunks<T>(ids: Id[], getter: (idsGroup: Id[]) => Prom
     return _.flatten(objsCollection);
 }
 
-export function promiseMap<T, S>(inputValues: T[], mapper: (value: T) => Promise<S>): Promise<S[]> {
-    const reducer = (acc$: Promise<S[]>, inputValue: T): Promise<S[]> =>
+export function promiseMap<T, S>(
+    inputValues: T[],
+    mapper: (value: T, idx: number) => Promise<S>
+): Promise<S[]> {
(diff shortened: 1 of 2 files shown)


tokland:
Moved to https://github.com/EyeSeeTea/dataset-configuration-scripts

PR #29, opened 2023-06-19
update events which met the condition

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/865cj18c2

### :fire: Testing

```bash
yarn start events update-events \
--url='http://USER:PASSWORD@HOST:PORT' \
--root-org-unit='root_org_unit_id' \
--event-ids='event_id_1,event_id_2,event_id_3' \
--data-element-id='data_element_id' \
--condition='true' \
--new-value='' \
--csv-path='./events.csv' \
--post
```


src/domain/usecases/UpdateEventDataValueUseCase.ts
```diff
@@ -0,0 +1,104 @@
...
+            logger.debug(`Events to change: ${eventsWithDvInCondition.length}`);
+            const stats = await this.eventRepository.saveAll(eventsWithDvInCondition);
+            return stats;
+        }
+
+        return {
+            created: 0,
+            ignored: 0,
+            updated: 0,
+        };
```
tokland, line 40: As a rule of thumb, favour full if/else (possible exception: when the if acts as guard). if (options.post) { ... } else { ... }

src/domain/usecases/UpdateEventDataValueUseCase.ts
```diff
@@ -0,0 +1,104 @@
+import _ from "lodash";
+import { createObjectCsvWriter } from "csv-writer";
```
tokland, line 2: Code in the domain cannot use external dependencies like csv-writer. Proposal: `EventExportSpreadsheetRepository implements EventExportRepository`

src/data/EventD2Repository.ts
```diff
@@ -0,0 +1,117 @@
...
+                    event: eventIds.join(";"),
+                    fields: eventsFields,
+                    totalPages: true,
+                    pageSize: 1e6,
+                } as EventsRequestWithFilter)
+                .getData()
+                .then(res => {
+                    const postEvents = eventIds.map(eventId => {
+                        const existingD2Event = res.events.find(d2Event => d2Event.event === eventId);
+                        const event = events.find(event => event.event === eventId);
```
tokland, line 65: Here is not a big problem because eventIds has a bounded size, but generally, create a indexed map (ie. _.keyby) to avoid O(n^2) (a find inside a map). `const event = eventsById[eventId]`.

src/data/EventD2Repository.ts
```diff
@@ -0,0 +1,117 @@
...
+                })
+                .catch(() => {
+                    logger.error(`Error getting events: ${eventIds.join(",")}`);
+                    return [];
+                });
+        });
+        return dataValues;
+    }
+
+    async saveAll(events: EventMetadata[]): Promise<Stats> {
```
tokland, line 51: Method a bit long, split in private methods/functions.

src/data/EventD2Repository.ts
```diff
@@ -0,0 +1,117 @@
...
+        const eventsIdsToSave = events.map(event => event.event);
+        const stats = await getInChunks<Stats>(eventsIdsToSave, async eventIds => {
```
tokland, line 53: [minor] `stats -> statsList` or something that signals that's a collection

src/domain/entities/Event.ts
```diff
@@ -0,0 +1,18 @@
...
+import { DataValue } from "./DataValue";
+
+export type EventMetadata = {
+    event: Id;
+    orgUnit: Id;
+    program: Id;
+    programStage: Id;
+    status: "ACTIVE" | "COMPLETED" | "VISITED" | "SCHEDULED" | "OVERDUE" | "SKIPPED";
+    eventDate: string;
+    dataValues: Pick<DataValue, "dataElement" | "value">[];
```
tokland, line 11: This `DataValue` refers to data set data values, which are not identical to program events. You can define the type directly here.

src/scripts/commands/events.ts
```diff
@@ -48,3 +51,57 @@ const moveOrgUnitCmd = command({
         if (!args.post) logger.info(`Add --post to update events`);
     },
 });
+
+const updateEventsDataValues = command({
+    name: "Update events",
+    description: "Update events which met the condition",
+    args: {
+        url: getApiUrlOption(),
+        eventId: option({
```
tokland, line 60: `eventIds`

src/domain/usecases/UpdateEventDataValueUseCase.ts
```diff
@@ -0,0 +1,104 @@
...
+            .filter(dv => dv.dataElement === options.dataElementId && dv.oldValue === options.condition)
+            .value();
+
+        await csvWriter.writeRecords(csvData);
+    }
+
+    private getEventsInCondition(events: EventMetadata[], options: MigrateOptions) {
+        const eventsWithDvInCondition = events
+            .map(event => {
+                const onlyDvInCondition = event.dataValues.map(dv => {
```
tokland, line 87: The name seems misleading, all dataValues are there (updated)

src/domain/usecases/UpdateEventDataValueUseCase.ts
```diff
@@ -0,0 +1,104 @@
...
+                    return {
+                        ...dv,
+                        value: dv.value === options.condition ? options.newValue : dv.value,
```
tokland, line 90: [important] We should be checking the dataElementId

src/data/EventD2Repository.ts
```diff
@@ -0,0 +1,117 @@
...
+                .then(res => {
+                    const postEvents = eventIds.map(eventId => {
+                        const existingD2Event = res.events.find(d2Event => d2Event.event === eventId);
+                        const event = events.find(event => event.event === eventId);
+                        if (!event) {
+                            throw Error("Cannot find event");
+                        }
+                        return {
+                            ...(existingD2Event || {}),
+                            ...event,
```
tokland, line 71: Caution: the analysis of completeness of properties must be done in `event.dataValues`. Example:

```
{
"lastUpdated": "2023-06-20T10:14:45.103",
"created": "2023-06-16T13:13:03.150",
"dataElement": "EZXB2m47njy",
"value": "200",
"providedElsewhere": false,
"createdByUserInfo": {
"uid": "SyI5Bx6M0Iv",
"firstName": "Adrian",
"surname": "Quintana",
"username": "adrian@eyeseetea.com"
},
"lastUpdatedByUserInfo": {
"uid": "SyI5Bx6M0Iv",
"firstName": "Adrian",
"surname": "Quintana",
"username": "adrian@eyeseetea.com"
}
}
```

I'd say we need to provide `providedElsewhere: boolean`  to the entity so is not lost on re-save.

tokland (approved):
I made some changes (feel free to challenge them, @eperedo !). All code good and it worked in my tests :+1:

PR #32, opened 2023-07-27
Monitor data values and send email

eperedo:
### :pushpin: References

### :art: Screenshots

Email example:

email

### :fire: Testing

d2-docker: [docker.eyeseetea.com/eyeseetea/dhis2-data:2.38-widp-dev](docker.eyeseetea.com/eyeseetea/dhis2-data:2.38-widp-dev)

Script can be executed using json or dhis datastore as storage:

#### Using json file:

```
yarn start datavalues monitoring-values \
--url='http://USER:PASSWORD@localhost:8080' \
--storage=json \
--settings-path=./settings.json \
--executions-path=./executions.json \
--email-path-template=./email.json \
--send-email-after-minutes=5
```

you can download an example of each json file in the clickup issue. 

#### Using dhis datastore:

For the following script you must create a `d2-notifications` namespace with a key **settings**. Put the content of the **settings.json** file from clickup in there and save it.

```
yarn start datavalues monitoring-values \
--url='http://USER:PASSWORD@localhost:8080' \
--storage=datastore \
--settings-path=d2-notifications.settings \
--executions-path=d2-notifications.executions \
--email-path-template=./email.json \
--send-email-after-minutes=5
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ Add option `--post` to actually (soft) delete the events.
 
 ### Email notification for data values
 
-Monitoring DataValues and sending an email depending on how long has passed since the last updated.
+Notify data values changes and sending an email depending on how long has passed since the last updated.
 
 Using json as a storage:
 
@@ -267,8 +267,8 @@ Using json as a storage:
 yarn start datavalues monitoring-values \
 --url='http://USER:PASSWORD@localhost:8080' \
 --storage=json \
---json-settings-path=./settings.json \
---json-executions-path=./executions.json \
+--settings-path=./settings.json \
+--executions-path=./executions.json \
 --email-path-template=./email.json \
 --send-email-after-minutes=5
 ```
@@ -279,7 +279,8 @@ Using dhis datastore as a storage:
 yarn start datavalues monitoring-values \
 --url='http://USER:PASSWORD@localhost:8080' \
 --storage=datastore \
---datastore-namespace=d2-notifications \
+--settings-path=d2-notifications.settings \
+--executions-path=d2-notifications.executions \
 --email-path-template=./email.json \
 --send-email-after-minutes=5
 ```
--- a/README.md
+++ b/README.md
@@ -257,6 +257,33 @@ $ yarn start programs get-duplicated-events \
 
 Add option `--post` to actually (soft) delete the events.
 
+### Email notification for data values
+
+Monitoring DataValues and sending an email depending on how long has passed since the last updated.
+
+Using json as a storage:
+
+```
+yarn start datavalues monitoring-values \
+--url='http://USER:PASSWORD@localhost:8080' \
+--storage=json \
+--json-settings-path=./settings.json \
+--json-executions-path=./executions.json \
+--email-path-template=./email.json \
+--send-email-after-minutes=5
+```
+
+Using dhis datastore as a storage:
+
+```
+yarn start datavalues monitoring-values \
+--url='http://USER:PASSWORD@localhost:8080' \
+--storage=datastore \
+--datastore-namespace=d2-notifications \
+--email-path-template=./email.json \
+--send-email-after-minutes=5
+```
+
 ## Notifications
 
 ### Send user info email
--- a/src/data/DataSetExecutionD2Repository.ts
+++ b/src/data/DataSetExecutionD2Repository.ts
@@ -1,0 +1,21 @@
+import { D2Api } from "@eyeseetea/d2-api/2.36";
+import { Async } from "domain/entities/Async";
+import { DataSetExecutionRepository } from "domain/repositories/DataSetExecutionRepository";
+import { NotificationStore } from "domain/usecases/SendNotificationDataValuesUseCase";
+import { SettingsOptions } from "domain/repositories/SettingsRepository";
+
+const key = "executions";
+
+export class DataSetExecutionD2Repository implements DataSetExecutionRepository {
+    constructor(private api: D2Api) {}
+
+    async save(value: NotificationStore, options: SettingsOptions): Async<void> {
+        const dataStore = this.api.dataStore(options.namespace);
+        await dataStore.save(key, value).response();
+    }
+
+    async get(options: SettingsOptions): Async<NotificationStore | undefined> {
+        const dataStore = this.api.dataStore(options.namespace);
+        return dataStore.get<NotificationStore>(key).getData();
+    }
+}
--- a/src/data/DataSetExecutionD2Repository.ts
+++ b/src/data/DataSetExecutionD2Repository.ts
@@ -1,21 +1,21 @@
-import { D2Api } from "@eyeseetea/d2-api/2.36";
+import { D2Api } from "types/d2-api";
 import { Async } from "domain/entities/Async";
 import { DataSetExecutionRepository } from "domain/repositories/DataSetExecutionRepository";
-import { NotificationStore } from "domain/usecases/SendNotificationDataValuesUseCase";
-import { SettingsOptions } from "domain/repositories/SettingsRepository";
-
-const key = "executions";
+import { DataSetExecution } from "domain/usecases/SendNotificationDataValuesUseCase";
+import { extractNameSpaceAndKeyFromPath, SettingsOptions } from "domain/entities/Settings";
 
 export class DataSetExecutionD2Repository implements DataSetExecutionRepository {
     constructor(private api: D2Api) {}
 
-    async save(value: NotificationStore, options: SettingsOptions): Async<void> {
-        const dataStore = this.api.dataStore(options.namespace);
+    async save(value: DataSetExecution, options: SettingsOptions): Async<void> {
+        const [namespace, key] = extractNameSpaceAndKeyFromPath(options.path);
+        const dataStore = this.api.dataStore(namespace);
         await dataStore.save(key, value).response();
     }
 
-    async get(options: SettingsOptions): Async<NotificationStore | undefined> {
-        const dataStore = this.api.dataStore(options.namespace);
-        return dataStore.get<NotificationStore>(key).getData();
+    async get(options: SettingsOptions): Async<DataSetExecution | undefined> {
+        const [namespace, key] = extractNameSpaceAndKeyFromPath(options.path);
+        const dataStore = this.api.dataStore(namespace);
+        return dataStore.get<DataSetExecution>(key).getData();
     }
 }
--- a/src/data/DataSetExecutionD2Repository.ts
+++ b/src/data/DataSetExecutionD2Repository.ts
@@ -2,19 +2,27 @@ import { D2Api } from "types/d2-api";
 import { Async } from "domain/entities/Async";
 import { DataSetExecutionRepository } from "domain/repositories/DataSetExecutionRepository";
 import { DataSetExecution } from "domain/usecases/SendNotificationDataValuesUseCase";
-import { extractNameSpaceAndKeyFromPath, SettingsOptions } from "domain/entities/Settings";
+import { SettingsOptions } from "domain/entities/Settings";
+
+export function extractNamespaceAndKeyFromPath(path: string): [string, string] {
+    const [namespace, key] = path.split(".");
+    if (!namespace || !key) {
+        throw Error(`Unable to get namespace and key from ${path}`);
+    }
+    return [namespace, key];
+}
 
 export class DataSetExecutionD2Repository implements DataSetExecutionRepository {
     constructor(private api: D2Api) {}
 
     async save(value: DataSetExecution, options: SettingsOptions): Async<void> {
-        const [namespace, key] = extractNameSpaceAndKeyFromPath(options.path);
+        const [namespace, key] = extractNamespaceAndKeyFromPath(options.path);
         const dataStore = this.api.dataStore(namespace);
         await dataStore.save(key, value).response();
     }
 
     async get(options: SettingsOptions): Async<DataSetExecution | undefined> {
-        const [namespace, key] = extractNameSpaceAndKeyFromPath(options.path);
+        const [namespace, key] = extractNamespaceAndKeyFromPath(options.path);
         const dataStore = this.api.dataStore(namespace);
         return dataStore.get<DataSetExecution>(key).getData();
     }
--- a/src/data/DataSetExecutionJsonRepository.ts
+++ b/src/data/DataSetExecutionJsonRepository.ts
@@ -1,0 +1,18 @@
+import fs from "fs";
+import { Async } from "domain/entities/Async";
+import { DataSetExecutionRepository } from "domain/repositories/DataSetExecutionRepository";
+import { NotificationStore } from "domain/usecases/SendNotificationDataValuesUseCase";
+import { SettingsOptions } from "domain/repositories/SettingsRepository";
+
+export class DataSetExecutionJsonRepository implements DataSetExecutionRepository {
+    async save(value: NotificationStore, options: SettingsOptions): Async<void> {
+        fs.writeFileSync(options.path, JSON.stringify(value, null, 4));
+    }
+
+    async get(options: SettingsOptions): Async<NotificationStore | undefined> {
+        const fileContent = fs.readFileSync(options.path, "utf-8");
+        if (!fileContent) return undefined;
+        const executionsJson = JSON.parse(fileContent);
+        return executionsJson;
+    }
+}
--- a/src/data/DataSetExecutionJsonRepository.ts
+++ b/src/data/DataSetExecutionJsonRepository.ts
@@ -1,17 +1,18 @@
 import fs from "fs";
 import { Async } from "domain/entities/Async";
 import { DataSetExecutionRepository } from "domain/repositories/DataSetExecutionRepository";
-import { NotificationStore } from "domain/usecases/SendNotificationDataValuesUseCase";
-import { SettingsOptions } from "domain/repositories/SettingsRepository";
+import { DataSetExecution } from "domain/usecases/SendNotificationDataValuesUseCase";
+import { SettingsOptions } from "domain/entities/Settings";
 
 export class DataSetExecutionJsonRepository implements DataSetExecutionRepository {
-    async save(value: NotificationStore, options: SettingsOptions): Async<void> {
+    async save(value: DataSetExecution, options: SettingsOptions): Async<void> {
         fs.writeFileSync(options.path, JSON.stringify(value, null, 4));
     }
 
-    async get(options: SettingsOptions): Async<NotificationStore | undefined> {
+    async get(options: SettingsOptions): Async<DataSetExecution | undefined> {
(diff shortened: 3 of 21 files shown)


src/domain/usecases/SendNotificationDataValuesUseCase.ts
```diff
@@ -0,0 +1,347 @@
...
+    emailPathTemplate: string;
+    sendEmailAfterMinutes: number;
+};
+
+export class SendNotificationDataValuesUseCase {
+    constructor(
+        private dataSetsRepository: DataSetsRepository,
+        private dataValuesRepository: DataValuesRepository,
+        private orgUnitRepository: OrgUnitRepository,
+        private dataSetStoreRepository: DataSetExecutionRepository,
```
tokland, line 45: Confusing: `dataSetStore/dataSetExecution`.

src/domain/usecases/SendNotificationDataValuesUseCase.ts
```diff
@@ -0,0 +1,347 @@
...
+    lastUpdated: string;
+};
+
+export type NotificationStore = Record<string, NotificationDetail>;
+
+type OptionsUseCase = {
+    storage: string;
+    jsonSettingsPath: string;
+    jsonExecutionsPath: string;
+    dataStoreNameSpace: string;
```
tokland, line 35: [clean arch] This is a bit complex, but the idea is: if we have chosen to hide the implementation (dataStore) behind a repo, no reference to the dataStore should appear in the domain (a use case, here). We have some generic "path" that's being interpreted in each repository (as a namespace/key, as a file path, whatever).

src/domain/usecases/SendNotificationDataValuesUseCase.ts
```diff
@@ -0,0 +1,347 @@
...
+
+        const dataSetExecutions = this.getDataSetsExecutions(
+            dataSetKeys,
+            settings,
+            dataSets,
+            orgUnits,
+            dataSetStore
+        );
+
+        if (_.isEmpty(dataSetStore)) {
```
tokland, line 84: `_.isEmpty(dataSetStore)` -> `!dataSetStore`

src/domain/usecases/SendNotificationDataValuesUseCase.ts
```diff
@@ -0,0 +1,347 @@
...
+
+    private async generateEmailTemplate(
+        dataSets: DataSet[],
+        dataSetId: string,
+        orgUnits: OrgUnit[],
+        orgUnitId: string,
+        dataValuesWithOldValue: DataValue[],
+        options: OptionsUseCase,
+        period: string | undefined
+    ) {
```
tokland, line 143: [best practices] Be explicit about the return type of a function, so a reader knows what's it's returning without hovering the method/inspecting the code. Also, if a bug is introduced, the error will show here and not on the caller code. [Same for all methods in the class]

src/domain/usecases/SendNotificationDataValuesUseCase.ts
```diff
@@ -0,0 +1,347 @@
...
+        const dataSetExecutions = keys.reduce<NotificationStore>((acum, key) => {
+            const storeInfo = dataSetStore ? dataSetStore[key] : undefined;
+            return {
+                ...acum,
+                [key]: {
+                    lastDateSentEmail: storeInfo?.lastDateSentEmail || "",
+                    lastUpdated: storeInfo?.lastUpdated || currentDate,
+                },
+            };
+        }, {});
```
tokland, line 233: This pattern `reduce + set by key` is more declarative and type-safe using lodash `fromPairs`.

src/domain/usecases/SendNotificationDataValuesUseCase.ts
```diff
@@ -0,0 +1,347 @@
...
+        execution.lastDateSentEmail = dateEmailSent;
+        execution.lastUpdated = dateEmailSent;
+        log.debug(`Email sent ${dateEmailSent}`);
+
+        await this.saveExecutions(
+            {
+                ...dataSetExecutions,
+                [key]: {
+                    ...execution,
+                },
```
tokland, line 293: `[key]: execution`

src/domain/usecases/SendNotificationDataValuesUseCase.ts
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    private async getUsersToNotify(dataValues: DataValue[], userGroup: string) {
+        const editorUsers = _(dataValues)
+            .map(dv => dv.storedBy)
+            .uniq()
+            .value();
+
+        const usersInGroup = await this.userRepository.getFromGroup([userGroup]);
+        return usersInGroup.filter(user => !editorUsers.includes(user.username));
```
tokland, line 324: Not a problem here, just a general note for bigger data structure: filter+includes is a nested O(n), O(m) loop, so O(n*m). Solution: Convert the first structure to O(1), for example, editorUsers can be a `Set`, the the includes (has) it's O(1) and the full loop is O(n)

src/domain/usecases/SendNotificationDataValuesUseCase.ts
```diff
@@ -0,0 +1,347 @@
...
+            if (lastDeUpdated) {
+                const currentDate = new Date();
+                const diffInMinutes = this.getDifferenceInMinutes(
+                    new Date(lastDeUpdated.lastUpdated),
+                    currentDate
+                );
+                return diffInMinutes;
+            }
+        }
+        return -1;
```
tokland, line 339: Magic values like `-1` are usually a red flag. Can we return `number | undefined`? (and change the caller accordingly)

src/scripts/commands/dataValues.ts
```diff
@@ -168,3 +175,83 @@ const postDanglingValuesCmd = command({
...
+        url: getApiUrlOption(),
+        storage: option({
+            type: oneOf(["datastore", "json"]),
+            long: "storage",
+            description: "datastore or json",
+        }),
+        dataStoreNameSpace: option({
+            type: string,
+            long: "datastore-namespace",
+            defaultValue: () => "namespace to be used ",
```
tokland, line 193: trailing space. I guess it's field "description", though

src/scripts/commands/dataValues.ts
```diff
@@ -168,3 +175,83 @@ const postDanglingValuesCmd = command({
...
+        }),
+        jsonSettingsPath: option({
+            type: string,
+            long: "json-settings-path",
+            description: "Path to json file that contains the dataset settings",
+            defaultValue: () => "",
```
tokland, line 199: [subjective] I prefer to set the optional al optional: `type: optional(string)` and remove a dubious empty string as default value.

src/scripts/commands/dataValues.ts
```diff
@@ -168,3 +175,83 @@ const postDanglingValuesCmd = command({
...
+        emailPathTemplate: option({
+            type: string,
+            long: "email-path-template",
+            description: "Path to the json file with email template information",
+        }),
+        sendEmailAfterMinutes: option({
+            type: number,
+            long: "send-email-after-minutes",
+            description: "Number of minutes to wait before sending the email notification. Default to 5",
+            defaultValue: () => 5,
```
tokland, line 216: However, here the defaultValue makes total sense. Can you move 5 to a variable to be both used in description and defaultValue?

src/domain/repositories/SettingsRepository.ts
```diff
@@ -0,0 +1,26 @@
...
+    orgUnit: Id;
+    period: string;
+};
+
+export type MonitoringConfig = {
+    userGroup: string;
+    monitoring: DataSetMonitoring[];
+};
+
+export type Settings = {
```
tokland, line 24: This type (and more in the file) should be moved to entity Settings.

src/data/DataSetExecutionJsonRepository.ts
```diff
@@ -0,0 +1,18 @@
...
+import { SettingsOptions } from "domain/repositories/SettingsRepository";
+
+export class DataSetExecutionJsonRepository implements DataSetExecutionRepository {
+    async save(value: NotificationStore, options: SettingsOptions): Async<void> {
+        fs.writeFileSync(options.path, JSON.stringify(value, null, 4));
+    }
+
+    async get(options: SettingsOptions): Async<NotificationStore | undefined> {
+        const fileContent = fs.readFileSync(options.path, "utf-8");
+        if (!fileContent) return undefined;
```
tokland, line 14: The executions file won't exist on first run and that should not be an error (now readFileSync fails)

src/data/DataSetExecutionJsonRepository.ts
```diff
@@ -0,0 +1,18 @@
...
+        const executionsJson = JSON.parse(fileContent);
+        return executionsJson;
```
tokland, line 16: Whenever you have extra time for this project, here it would be good to use purify-ts Codec (we've done it in other projects), to validate data from JSON (files, dataStore, ...)

src/domain/usecases/SendNotificationDataValuesUseCase.ts
```diff
@@ -0,0 +1,347 @@
...
+    private async sendEmailToUsersInGroup(users: User[], subject: string, body: string) {
+        await promiseMap(users, async user => {
+            await this.notificationsRepository.send({
+                attachments: [],
+                subject,
+                body: {
+                    contents: body,
+                    type: "html",
+                },
+                recipients: [user.email],
```
tokland, line 308: Why a loop instead of using the array?

src/domain/usecases/SendNotificationDataValuesUseCase.ts
tokland, line 324: In general, can you add more debug messages for all this intermediate variables? This way the caller will see the logic behind emails being sent or not.

src/domain/usecases/SendNotificationDataValuesUseCase.ts
```diff
@@ -0,0 +1,347 @@
...
+    subject: string;
+    body: string;
+};
+
+export type NotificationDetail = {
+    lastDateSentEmail: string;
+    lastUpdated: string;
+};
+
+export type NotificationStore = Record<string, NotificationDetail>;
```
tokland, line 29: Not sure we need to store the dataValue we have notified. As I see it, we get the dataValues from the last notification marker, get the ones which match the time window, and update the marker.

src/data/UserD2Repository.ts
```diff
@@ -196,4 +196,59 @@ export class UserD2Repository implements UserRepository {
...
+                    filter: {
+                        identifiable: {
+                            in: values,
+                        },
+                    },
+                },
+            })
+            .getData();
+
+        const userGroup = response.userGroups[0];
```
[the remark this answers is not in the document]
eperedo: Since the goal is to get an array of users from a group, can I rename the method to `getFromGroupByIdentifiable(value: Identifiable): Async<User[]>` and keep the same logic?

src/data/UserD2Repository.ts
```diff
@@ -196,4 +196,59 @@ export class UserD2Repository implements UserRepository {
...
+        if (userGroup) {
+            const ids = userGroup.users.map(user => user.id);
+            const users = await promiseMap(_(ids).chunk(50).value(), async userIds => {
+                const response = await this.api.metadata
```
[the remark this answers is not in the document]
eperedo: `/api/userGroups` endpoint does not return the email

https://play.dhis2.org/2.37.9.1/api/userGroups/wl5cDMuUhmF.json?fields=id,name,users[*]

```
{
  "name": "dataset-notification",
  "id": "hgJiO8evtcJ",
  "users": [
    {
      "displayName": "user dev",
      "name": "user dev",
      "id": "smGarTiKDdV",
      "username": "dev.user"
    }
  ]
}
```

src/domain/repositories/SettingsRepository.ts
```diff
@@ -0,0 +1,26 @@
...
+    dataSets: Record<string, MonitoringConfig>;
```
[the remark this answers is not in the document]
eperedo: we already had a config used by the client 

```json
[
  {
    "monitoring": false,
    "orgUnit": "hmZE3mVAZFf",
    "period": "2021"
  },
  {
    "monitoring": false,
    "orgUnit": "oofyLUJJ6Vy",
    "period": "2021"
  },
  {
    "monitoring": false,
    "orgUnit": "Nrnay4JA8Ga",
    "period": "2021"
  }
]
```

wrapped that json inside a `dataSets` key to keep settings open to other set of notifications.

```json
{
  "other-type-notification-here": { ... },
  "dataSets": {
    "MAL - WMR Form": {
      "monitoring": [
        {
          "enable": true,
          "orgUnit": "BFA",
          "period": "2021"
        },
        {
          "enable": true,
          "orgUnit": "AFG",
          "period": "2021"
        },
        {
          "enable": true,
          "orgUnit": "BLZ",
          "period": "2021"
        }
      ],
      "userGroup": "hgJiO8evtcJ"
    }
  }
}
```

src/domain/usecases/SendNotificationDataValuesUseCase.ts
eperedo: I don't get it. Here `lastUpdated` is being used to save the date of the last dataValue notified. should I save that value in a different format or attribute?

src/data/UserD2Repository.ts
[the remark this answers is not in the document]
tokland: Yes, that would work

src/data/UserD2Repository.ts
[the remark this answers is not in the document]
tokland: You're right, that changed on latest versions, so nothing to change.

src/domain/repositories/SettingsRepository.ts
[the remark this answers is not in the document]
tokland: Ok, if Settings is only going to be used for the notifications, then that's ok.

README.md
```diff
@@ -259,16 +259,16 @@ Add option `--post` to actually (soft) delete the events.
 
 ### Email notification for data values
 
-Monitoring DataValues and sending an email depending on how long has passed since the last updated.
+Notify data values changes and sending an email depending on how long has passed since the last updated.
```
tokland, line 262: `and sending` -> `and send`

src/data/DataSetExecutionD2Repository.ts
```diff
@@ -1,21 +1,21 @@
-import { D2Api } from "@eyeseetea/d2-api/2.36";
+import { D2Api } from "types/d2-api";
 import { Async } from "domain/entities/Async";
 import { DataSetExecutionRepository } from "domain/repositories/DataSetExecutionRepository";
-import { NotificationStore } from "domain/usecases/SendNotificationDataValuesUseCase";
-import { SettingsOptions } from "domain/repositories/SettingsRepository";
-
-const key = "executions";
+import { DataSetExecution } from "domain/usecases/SendNotificationDataValuesUseCase";
+import { extractNameSpaceAndKeyFromPath, SettingsOptions } from "domain/entities/Settings";
```
tokland, line 5: Naming: `NameSpace` -> `Namespace` (DHIS2 casing)

tokland, line 5: The entity knows nothing about dataStore namespace/key, that's a dhis2 detail, so the function should be implemented in this data repo.

src/data/UserD2Repository.ts
```diff
@@ -214,9 +214,13 @@ export class UserD2Repository implements UserRepository {
             })
             .getData();
 
-        const userGroup = response.userGroups[0];
-        if (userGroup) {
-            const ids = userGroup.users.map(user => user.id);
+        const allUsers = _(response.userGroups)
+            .map(ug => ug.users)
+            .flatten()
```
tokland, line 219: map + flatten -> flatMap

src/data/UserD2Repository.ts
```diff
@@ -214,9 +214,13 @@ export class UserD2Repository implements UserRepository {
...
+            .value();
+
+        if (allUsers.length > 0) {
```
tokland, line 222: [subjective] Typically, the "guard style" is `if (allUsers.length === 0) return []`

src/data/UserD2Repository.ts
```diff
@@ -214,9 +214,13 @@ export class UserD2Repository implements UserRepository {
...
+            const ids = allUsers.map(user => user.id);
             const users = await promiseMap(_(ids).chunk(50).value(), async userIds => {
```
tokland, line 224: [note, subjective] typically when calling a single method: `_.chunk(ids, 50)`

src/scripts/commands/dataValues.ts
```diff
@@ -179,30 +181,23 @@ const postDanglingValuesCmd = command({
 const monitoringDataValues = command({
     name: "monitoring-values",
     description:
-        "Monitoring data values and sending an email depending on how long has passed since the last Updated",
+        "Notify data value changes and sending an email depending on how long has passed since the last updated",
```
tokland, line 184: and send

src/domain/usecases/SendNotificationDataValuesUseCase.ts
```diff
@@ -42,35 +41,18 @@ export class SendNotificationDataValuesUseCase {
...
-        const orgUnits = await this.orgUnitRepository.get(orgUnitIds);
-        const dataSetStore = await this.dataSetStoreRepository.get({
-            path: options.jsonExecutionsPath,
-            namespace: options.dataStoreNameSpace,
+        const { dataSets, orgUnits } = await this.getOrgUnitAndDataSets(dataSetKeys, settings);
+        const dataSetStore = await this.dataSetExecutionRepository.get({
+            path: options.executionsPath,
         });
 
         const dataSetExecutions = this.getDataSetsExecutions(
```
tokland, line 58: dataSetKeys can be extracted from settings, no need to pass it.

src/domain/usecases/SendNotificationDataValuesUseCase.ts
```diff
@@ -42,35 +41,18 @@ export class SendNotificationDataValuesUseCase {
...
-            .flatten()
-            .value();
-
-        const dataSets = await this.dataSetsRepository.getBy(dataSetKeys);
-        const orgUnits = await this.orgUnitRepository.get(orgUnitIds);
-        const dataSetStore = await this.dataSetStoreRepository.get({
-            path: options.jsonExecutionsPath,
-            namespace: options.dataStoreNameSpace,
+        const { dataSets, orgUnits } = await this.getOrgUnitAndDataSets(dataSetKeys, settings);
+        const dataSetStore = await this.dataSetExecutionRepository.get({
```
tokland, line 54: dataSetStore -> dataSetExecution ?

tokland (approved):
[code-only review] Thanks, Eduardo! All the notes in this PR are very minor and don't interfere with the feature at all, so I'll approve the PR.

PR #33, opened 2023-08-28
fix: send email to each user in the group

eperedo:
### :pushpin: References

### :art: Screenshots

Unique emails for each user

### email: eduardo@eyeseetea.com
image

### email: jrperedo@gmail.com
image

### DIFF
--- a/src/domain/usecases/SendNotificationDataValuesUseCase.ts
+++ b/src/domain/usecases/SendNotificationDataValuesUseCase.ts
@@ -362,14 +362,16 @@ export class SendNotificationDataValuesUseCase {
     }
 
     private async sendEmailToUsersInGroup(users: User[], subject: string, body: string): Async<void> {
-        await this.notificationsRepository.send({
-            attachments: [],
-            subject,
-            body: {
-                contents: body,
-                type: "html",
-            },
-            recipients: users.map(user => user.email),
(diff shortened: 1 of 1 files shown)


src/domain/usecases/SendNotificationDataValuesUseCase.ts
```diff
@@ -362,14 +362,16 @@ export class SendNotificationDataValuesUseCase {
...
+        await promiseMap(users, async user => {
+            await this.notificationsRepository.send({
+                attachments: [],
+                subject,
+                body: {
+                    contents: body,
+                    type: "html",
+                },
+                recipients: [user.email],
+            });
```
tokland, line 374: So finally sending N emails instead of BCC, right?

tokland (approved):

src/domain/usecases/SendNotificationDataValuesUseCase.ts
eperedo: Yep, bcc is still including info from other users (gmail and outlook, not sure about other mail clients).

image

tokland: Oh, good to know, it looks like a bug in nodemailer. Sending N emails seems also fine for this use case.

PR #35, opened 2023-08-31
add timezone support

eperedo:
### :pushpin: References

### :memo: Implementation

We were having problems related to the time zone of the server. In dev time zone is UTC, but in PREPROD (and maybe PROD) is Europe/Madrid and it was generating weird responses from the `/api/dataValueSets` API. So in order to avoid future errors the script now get the time zone from the server `/api/system/info`. 

The problem is that [d2-api does not expose the type for the property `serverTimeZoneId`](https://github.com/EyeSeeTea/d2-api/blob/development/src/api/system.ts#L160). Just a typing issue so I'm disabling type checking in the current implementation.

Also I'm adding a new parameter for the script `--timezone` [because of this](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-238/maintenance.html#:~:text=If%20the%20user%20requesting%20this,considered%20sensitive%20will%20be%20included.):

```
If the user requesting this resource does not have full authority 
then only properties which are not considered sensitive will be included.
```

so we have an alternative to execute the script even if the user does not have full permissions.

```
LOG_LEVEL=debug yarn start datavalues monitoring-values 
--url='http://user:password@localhost:8080' 
--storage=datastore 
--settings-path=d2-reports.monitoring 
--executions-path=d2-reports.executions 
--email-path-template=./email.json 
--send-email-after-minutes=1 
--timezone=Europe/Madrid
```

After pulling changes please install dependencies using `yarn` or `yarn install`


MiquelAdell (approved):
Tested, it works, thanks a lot @eperedo

PR #38, opened 2023-10-09
move attributes from a program

eperedo:
### :pushpin: References

* **Issue:** Closes https://app.clickup.com/t/865d6wugn

### :memo: Implementation

Only sending the attributes information in the POST request because if the program has `programStages` with `dataElements` that were removed saving the whole TEI generates errors like this:

```
DataElement `Jum2ZQV1RDQ` is not part of `NtbHvjQEW3n` program stage
```

### :fire: Notes for the reviewer

**metadata:** attached in the issue
**docker:** docker.eyeseetea.com/samaritans/dhis2-data:2.36.10.1-sp-cpr

Example:

```bash
yarn start programs move-attribute \
--url='http://USER:PASSWORD@HOST:PORT' \
--program-id=WCJhvPcJomX \
--from-attribute-id=MyOceAlOxLK \
--to-attribute-id=YqcpwxxRc1D
```

### :video_camera: Screenshots/Screen capture

Completed enrollment
image

Open enrollment
image


src/data/ProgramsD2Repository.ts
```diff
@@ -87,20 +101,109 @@ export class ProgramsD2Repository implements ProgramsRepository {
         return d2ProgramRules.run(options);
     }
 
+    async getAll(options: MoveProgramAttributeOptions): Async<ProgramAttributes[]> {
+        const trackedEntities = await this.getFromTracker<D2TrackedEntity>("trackedEntities", {
```
tokland, line 105: We were using manual calls to the new tracker, but at this point d2-api already should support see (https://github.com/EyeSeeTea/d2-api/pull/135)  from v1.14.0. Check with @adrianq if we do this change now or we'll start using it on future developments.

src/data/ProgramsD2Repository.ts
```diff
@@ -87,20 +101,109 @@ export class ProgramsD2Repository implements ProgramsRepository {
...
+                        value: attribute.value,
+                        storedBy: attribute.storedBy,
+                    };
+                }),
+                programId: options.programId,
+            };
+        });
+    }
+
+    async saveAttributes(programs: ProgramAttributes[]): Async<Stats> {
```
tokland, line 127: For me, `ProgramAttributes` is a confusing name. It's a TrackedEntityInstance with some basic fields, including attributes. We don't a TEI entity yet. I'd create it (TrackedEntity.ts, as the new version of DHIS drop the "Instance"), and move this code to a new repository for this entity. What do you think?

src/data/ProgramsD2Repository.ts
```diff
@@ -87,20 +101,109 @@ export class ProgramsD2Repository implements ProgramsRepository {
...
+            const trackedEntities = await this.getFromTracker<D2TrackedEntity>("trackedEntities", {
+                orgUnitIds: undefined,
+                programIds: programsIds,
+                trackedEntity: teiIds.join(";"),
+            });
+
+            const teisToSave = trackedEntities.map(tei => {
+                const currentProgram = programsByKey[tei.trackedEntity];
+                if (!currentProgram) throw Error(`Cannot find program: ${tei.trackedEntity}`);
+                const attributes = currentProgram?.attributes || tei.attributes;
```
tokland, line 150: There are two `currentProgram?` but we have a guard for the undefined value (which is correct), so the code can be simplified.

src/data/ProgramsD2Repository.ts
```diff
@@ -87,20 +101,109 @@ export class ProgramsD2Repository implements ProgramsRepository {
...
+        const stats = await getInChunks<Stats>(teisToFetch, async teiIds => {
+            const trackedEntities = await this.getFromTracker<D2TrackedEntity>("trackedEntities", {
+                orgUnitIds: undefined,
+                programIds: programsIds,
+                trackedEntity: teiIds.join(";"),
+            });
+
+            const teisToSave = trackedEntities.map(tei => {
+                const currentProgram = programsByKey[tei.trackedEntity];
+                if (!currentProgram) throw Error(`Cannot find program: ${tei.trackedEntity}`);
```
tokland, line 149: `Cannot find tracked entity instance`

src/domain/entities/ProgramAttributes.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export type ProgramAttributes = {
+    id: Id;
+    attributes: Attribute[];
+    orgUnit: Id;
+    trackedEntityType: Id;
+    programId: Id;
+};
+
+export type Attribute = {
```
tokland, line 11: let's call that AttributeValue to differentiate the attribute itself from the value in a TEI.

src/domain/repositories/ProgramsRepository.ts
```diff
@@ -1,12 +1,16 @@
...
 import { Timestamp } from "domain/entities/Date";
+import { ProgramAttributes } from "domain/entities/ProgramAttributes";
 import { ProgramExport } from "domain/entities/ProgramExport";
+import { Stats } from "domain/entities/Stats";
 
 export interface ProgramsRepository {
     export(options: { ids: Id[] }): Async<ProgramExport>;
     import(programExport: ProgramExport): Async<void>;
     runRules(options: RunRulesOptions): Async<void>;
+    getAll(options: MoveProgramAttributeOptions): Async<ProgramAttributes[]>;
```
tokland, line 12: It's weird that a `getAll` method has options `MoveProgramAttributeOptions`

src/domain/usecases/MoveProgramAttributeUseCase.ts
```diff
@@ -0,0 +1,68 @@
...
+                const fromAttribute = tei.attributes.find(attr => attr.attribute === options.fromAttributeId);
+                if (!fromAttribute) return undefined;
+                const toAttribute = tei.attributes.find(attr => attr.attribute === options.toAttributeId);
+
+                return {
+                    ...tei,
+                    attributes: toAttribute
+                        ? this.generateAttributes(tei.attributes, fromAttribute, toAttribute)
+                        : [
+                              ...this.generateAttributes(tei.attributes, fromAttribute, toAttribute),
```
tokland, line 25: this call is identical is both branches, so it can be a const before the return

src/data/ProgramsD2Repository.ts
eperedo: I tried to update but there was some errors in `ProgramEventsD2Repository`. @adrianq told me we can postpone the migration since we will need to do some test in other scripts not related with the current one.

src/data/ProgramsD2Repository.ts
```diff
@@ -101,92 +96,9 @@ export class ProgramsD2Repository implements ProgramsRepository {
...
-                ignored: acum.ignored + stat.ignored,
-                updated: acum.updated + stat.updated,
-            };
-        }, initialStats);
-    }
-
     /* Private */
 
-    private async postTracker(key: TrackerDataKey, objects: object[]): Async<TrackerResponse[]> {
+    async postTracker(key: TrackerDataKey, objects: object[]): Async<TrackerResponse[]> {
```
tokland, line 101: I see you made these methods public (and getFromTracker) so they can be used from the other repo. Typically, we would create a separate file (for example: D2Tracker.ts, with a class D2Tracker) that encapsulate this logic.

src/data/TrackedEntityD2Repository.ts
```diff
@@ -0,0 +1,119 @@
...
+import { getInChunks } from "./dhis2-utils";
+import { Stats } from "domain/entities/Stats";
+import {
+    TrackedEntityFilterParams,
+    TrackedEntityRepository,
+} from "domain/repositories/TrackedEntityRepository";
+import { TrackedEntity } from "domain/entities/TrackedEntity";
+import { D2TrackedEntity, ProgramsD2Repository } from "./ProgramsD2Repository";
+
+const initialStats = {
```
tokland, line 14: `const initialStats: Stats`. Pros: 1) This way you get validation here (and not in usage), 2) On refactor (for example a field renaming), TS will know that you are here defining a Stat and act accodingly.

src/data/TrackedEntityD2Repository.ts
```diff
@@ -0,0 +1,119 @@
...
+    recordsSkipped: [],
+    errorMessage: "",
+    created: 0,
+    ignored: 0,
+    updated: 0,
+};
```
tokland, line 20: This object is repeated in 2 repos. Also, we have some logic, `stats.reduce`.  Conclusion: Let's create a class Stats, with all we need (empty value, as static method, the accumulation)

src/data/TrackedEntityD2Repository.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export class TrackedEntityD2Repository implements TrackedEntityRepository {
+    private programsD2Repository: ProgramsD2Repository;
+
+    constructor(private api: D2Api) {
+        this.programsD2Repository = new ProgramsD2Repository(this.api);
```
tokland, line 26: And then we don't instantiate a repo in a another repo.

src/data/TrackedEntityD2Repository.ts
```diff
@@ -0,0 +1,119 @@
...
+                    };
+                }),
+                programId: params.programId,
+            };
+        });
+    }
+
+    async saveAttributes(teis: TrackedEntity[]): Async<Stats> {
+        if (teis.length === 0)
+            return { created: 0, ignored: 0, updated: 0, errorMessage: "", recordsSkipped: [] };
```
tokland, line 57: This is an empty stat, right? eventually we can use `Stats.empty()`

src/domain/repositories/TrackedEntityRepository.ts
```diff
@@ -0,0 +1,15 @@
+import { Async } from "domain/entities/Async";
+import { Id } from "domain/entities/Base";
+import { Stats } from "domain/entities/Stats";
+import { TrackedEntity } from "domain/entities/TrackedEntity";
+
+export interface TrackedEntityRepository {
+    getAll(params: TrackedEntityFilterParams): Async<TrackedEntity[]>;
+    saveAttributes(programs: TrackedEntity[]): Async<Stats>;
```
tokland, line 8: saveAttributes -> save, programs -> trackedEntities

src/data/D2Tracker.ts
```diff
@@ -0,0 +1,111 @@
...
+                { [key]: objectsGroup },
+                { payloadId: `${key}-${initial}` }
+            );
+            result.push(trackerResponse);
+            page++;
+        }
+        return result;
+    }
+
+    async postTrackerData(data: object, options: { payloadId: string }): Async<TrackerResponse> {
```
tokland: private

src/data/UserD2Repository.ts
```diff
@@ -150,24 +150,7 @@ export class UserD2Repository implements UserRepository {
...
-            },
-            {
-                recordsSkipped: [],
-                errorMessage: "",
-                created: 0,
-                ignored: 0,
-                updated: 0,
-            }
-        );
+        return Stats.combine(stats);
```
tokland, line 153: nice!

src/domain/entities/Stats.ts
```diff
@@ -1,9 +1,47 @@
...
+    public readonly ignored: number;
+    public readonly updated: number;
+    public readonly errorMessage: string;
+
+    constructor(attrs: StatsAttrs) {
+        this.recordsSkipped = attrs.recordsSkipped;
+        this.created = attrs.created;
+        this.ignored = attrs.ignored;
+        this.updated = attrs.updated;
+        this.errorMessage = attrs.errorMessage;
```
tokland, line 23: To avoid this duplication we have Struct in the new skeleton, but that's perfectly fine.

tokland (approved):
All good, thx! I've just added a private for a method in there, not important.

PR #42, opened 2023-12-28
Feature/new user permisions cmd

idelcano:
### :pushpin: References

* **Issue:** Closes https://app.clickup.com/t/3x19q55

### :memo: Implementation

The script checks that all users have the appropriate user roles. In WIDP, we have some template users linked to user groups, for example: widpit.template with the WIDP IT group. Any user assigned to the WIDP IT group can have at most all the authorities that the roles of the widpit.template user have, not one more.

The program goes through all the users looking for users without a widp template group, and if there isn't one, it assigns the lowest level one because it is mandatory that users have one of those roles. After correcting the groups if necessary, it downloads (updated) and goes through all the users again, looking for invalid roles; if there are any, a copy of the user with the corrected roles is made. In the end, all users who had to remove any role are sent, and a report is sent to the instance with: The number of users who needed to be added to the lowest-level template group, and the list of their usernames and IDs. The number of users who had roles removed, and their list of usernames and IDs, plus two JSON files, one with those users as they were on the server before running the script, and another afterward, to facilitate their analysis if there were any problems and to audit all changes.

The script needs a config.json file with the URL, username, and password, and to read from the datastore d2-tools->user-monitoring all the possible parameters and metrics to function, including the list of groups and users that are templates, which is the lowest level, and exclusions to the checking.

### :fire: Notes for the reviewer

fill config.json file.
Copy d2-tools -> user-monitoring datastore.
i will program a meet with  @eperedo to explain.

```
yarn build:dev
LOG_LEVEL=debug node --inspect-brk dist/index.js usermonitoring run-users-monitoring   --config-file config.json
```


src/domain/repositories/UserAuthoritiesRepository.ts
```diff
@@ -0,0 +1,20 @@
+import { User } from "data/d2-users/D2Users.types";
+import {
+    TemplateGroupWithAuthorities,
+    UserMonitoringCountResponse,
+    UserMonitoringDetails,
+    UsersOptions,
+} from "domain/entities/UserMonitoring";
+
+export interface UserAuthoritiesRepository {
```
eperedo, line 9: [Clean Arc] Normally we want our repos to be very generic so we can't reuse them between different usecases.
for this particular case we have a mix of application logic and data layer code.
example: `processUserGroups`

Data layer code:

```ts
private async getGroups(groupsIds: string[]): Promise<UserGroup[]> {}

async pushUsersToGroup(minimalUserGroup: UserGroup[], userIdWithoutGroups: IdItem[]): Promise<string> {}
```
        
Application logic must be at the entity level (usecase/domain)

```ts
const response = await this.addLowLevelTemplateGroupToUsersWithoutAny(
  completeTemplateGroups,
  allUsersGroupCheck,
  minimalGroupId
 );
```

```ts
const userIdWithoutGroups: Item[] = this.detectUserIdsWithoutGroups(
            completeTemplateGroups,
            allUsersGroupCheck,
            minimalGroupId
);
 ```

```ts
  this.pushUsersWithoutGroupsWithLowLevelGroup(userIdWithoutGroups, minimalGroupId);
```
        
In order to have a better separation we can create a `UserGroupRepository` that we can use to get/save UserGroup information.

```ts
export interface UserGroupRepository {
  getByIds(ids: Id[]): Promise<UserGroup[]>;
  save(userGroup: UserGroup): Promise<string>;
}
```

 and move all the `processUserGroups` logic to the `RunUserMonitoringUseCase`:

```ts
export class RunUserMonitoringUseCase {
            constructor(
                private userGroupRepository: UserGroupRepository,
                ...
            ) {}
            
            execute(options: UsersOptions) {
                ...

                // get the minimal group information
                const groups = await this.userGroupRepository.getByIds([options.minimalGroupId.id]);
                const minimalGroup = _(groups).first();
                if (!minimalGroup) throw Error(`Cannot find minimal group ${options.minimalGroupId.id}`);
                // from here we can call all the other methods and directly use the minimalGroup variable
                const lowLevelTemplateGroup = await this.addLowLevelTemplateGroupToUsersWithoutAny(
                    completeTemplateGroups,
                    allUsersGroupCheck,
                    minimalGroup 
                ); 

                const userIdWithoutGroups = this.detectUserIdsWithoutGroups(
                    completeTemplateGroups,
                    allUsersGroupCheck,
                    minimalGroup.id
                );

                // instead saving this method only generates the values for invalidUsersCount, listOfAffectedUsers and the minimalUserGroup to save
                const groupsWithoutLowLevelGroup = this.pushUsersWithoutGroupsWithLowLevelGroup(userIdWithoutGroups, minimalGroup);

                // here we save the minigal group
                await this.userGroupRepository.save(groupsWithoutLowLevelGroup.minimalUserGroup)
            }
```

with these changes now we have all the application logic in the use case and we have all of the DHIS2 implementation details in the `UserGroupRepository` and if we need to get/save usergroups in other usecase we can totally reuse this repository without breaking any functionality here.

Using this repository as an example, but this apply for all the other repositories.

src/domain/repositories/UserAuthoritiesRepository.ts
```diff
@@ -0,0 +1,20 @@
...
+    processUserGroups(
+        options: UsersOptions,
+        completeTemplateGroups: TemplateGroupWithAuthorities[],
+        users: User[]
+    ): Promise<UserMonitoringCountResponse>;
+    processUserRoles(
```
eperedo, line 15: [Clean Arc.] Same as before, for the processUserRoles we can extract all the logic into the use case, build the users we want to udpate and then in the usecase call the `UserRepository.save(userToPost);`

src/domain/repositories/UserMonitoringMetadataRepository.ts
```diff
@@ -0,0 +1,68 @@
+import { ProgramMetadata, UserRes, UserRoleAuthority, User } from "data/d2-users/D2Users.types";
+import { Async } from "domain/entities/Async";
+
+export interface UserMonitoringMetadataRepository {
+    getTemplateAuthorities(options: UsersOptions): Promise<Async<TemplateGroupWithAuthorities[]>>;
```
eperedo, line 5: [Clean Arc.] Same process, I have identify that you will need a `UserRoleRepository` with methods `getAll` and `getByIds` and UserRepository with a method `getByIds`

src/domain/repositories/UserMonitoringMetadataRepository.ts
```diff
@@ -0,0 +1,68 @@
...
+    getMetadata(programId: string): Promise<Async<ProgramMetadata>>;
+    getAllUsers(excludeIds: string[], exclude?: boolean): Promise<Async<User[]>>;
+}
+
+export interface UserMonitoringDetails extends UserMonitoringCountResponse {
```
eperedo, line 10: There are some interfaces here that are the same in `UserMonitoring.ts`, can we reuse them?

eperedo (changes requested):
Hi @idelcano thanks for the code. The script seems to work great on my local instance:

image

I have added some observations and a proposal regarding clean arch. Let me know if you have any questions:

src/data/UserAuthoritiesD2Repository.ts
```diff
@@ -0,0 +1,396 @@
...
+    ): Item[] {
+        return _.compact(
+            allUsersGroupCheck.map(user => {
+                const templateGroupMatch = completeTemplateGroups.find(template => {
+                    return user.userGroups.some(
+                        userGroup => userGroup != undefined && template.group.id == userGroup.id
+                    );
+                });
+
+                if (templateGroupMatch == undefined) {
```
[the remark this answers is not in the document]
idelcano: I didnt change it due if the template is undefined we need do some actions, if i change the code by the suggestion the logic changes

ifoche:
This PR started long time ago. Eduardo and Ina went through it and there are some pending modifications that were requested by Eduardo. Since this started some other additional features have been created in separated branches by Ina and Nazar. Ina is implementing the changes requested by Eduardo in his separate branch. Nazar will take those changes into his separate branch too, and only after that, this PR will be discarded (assuming that all these recommendations will be implemented there).

@eperedo to be aware of all this, as well as @nshandra and @idelcano

PR #50, opened 2024-06-10
Feature/ignore invalid users

idelcano:
Added new boolean variable in the datastore to avoid the push of the report without fix before the usergroups.

The script assume that the user has a valid usergroup adding the minimal usergroup to that users without any valid usergroup, to allow to the report continue.

forceMinimalGroupForUsersWithoutGroup

### DIFF
--- a/README.md
+++ b/README.md
@@ -448,9 +448,9 @@ A config json file to get the user/password and server:
 
 d2-tools -> two-factor-monitoring:
 
-A program id with the id of the program in dhis
+A push program variable with the id of the program in dhis
 
-A group id to filter the users that should have two factor activated
+A two factor group to filter the users that should have two factor activated
 
 The datastore must contain:
 
@@ -486,19 +486,19 @@ List of excluded users (the script will ignore them) (could be an empty list [])
 A minimum group to add that group to the users without any template group (WIDP requirement)
 
 A minimum role to add that role to the users without any role (DHIS requirement)
-A push_program_id variable with the program ID to send the report
+A pushProgram variable with the program ID to send the report
 
 A list of templates (user with the valid roles, and group to identify which users could use those roles).
 
 A list with flags to control the script (permissionFixerConfig).
 
-A boolean variable push_report to determine if the script must send the report after processing all users
+The pushReport boolean variable determines whether the script should send the report after processing all users.
 
-A boolean variable pushFixedUserGroups to allow push the fixed usergroups (for those users without widp template user group)
+The pushFixedUserGroups boolean variable enables pushing fixed user groups for users who do not have a WIDP template user group.
 
-A boolean variable pushFixedUsersRoles to allow the push of the fixed users.
+The pushFixedUsersRoles boolean variable allows the pushing of fixed user roles.
 
-A boolean variable forceMinimalGroupForUsersWithoutGroup to determine ignore the users with invalid usergroup exception, assuming for that users the minimal group
+The forceMinimalGroupForUsersWithoutGroup boolean variable determines how to handle users with invalid user groups. If set to true, it treats these users as if they are assigned to the minimal group without updating their user group. If set to false, it assigns the minimal group to users without a template user group on the server.
 
 An example of the datastore:
 
--- a/README.md
+++ b/README.md
@@ -486,14 +486,19 @@ List of excluded users (the script will ignore them) (could be an empty list [])
 A minimum group to add that group to the users without any template group (WIDP requirement)
 
 A minimum role to add that role to the users without any role (DHIS requirement)
-
-A boolean variable push_report to determine if the script must send the report after processing all users
-
 A push_program_id variable with the program ID to send the report
 
 A list of templates (user with the valid roles, and group to identify which users could use those roles).
 
 A list with flags to control the script (permissionFixerConfig).
+
+A boolean variable push_report to determine if the script must send the report after processing all users
+
+A boolean variable pushFixedUserGroups to allow push the fixed usergroups (for those users without widp template user group)
+
+A boolean variable pushFixedUsersRoles to allow the push of the fixed users.
+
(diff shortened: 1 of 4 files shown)


src/domain/usecases/user-monitoring/permission-fixer/RunUserPermissionUseCase.ts
```diff
@@ -322,6 +332,24 @@ export class RunUserPermissionUseCase {
...
+                    userGroup => userGroup != undefined && template.group.id == userGroup.id
+                );
+            });
+            if (templateGroupMatch == undefined) {
+                //template not found -> all roles are invalid except the minimal role
+                user.userGroups.push({ id: minimalGroupId, name: "Minimal Group" });
+            }
+        });
+    }
+
```
nshandra, line 352: A return-less map is a `forEach`, in this case its better to use a functional approach, that way its clear that we are modifying allUsers. 
```typescript
private addMinimalUserGroupToUsersWithoutUserGroup(
    allUsers: PermissionFixerUser[],
    completeTemplateGroups: PermissionFixerTemplateGroupExtended[],
    minimalGroupId: string
) {
    return allUsers.map(user => {
        const templateGroupMatch = completeTemplateGroups.find(template => {
            return user.userGroups.some(
                userGroup => userGroup != undefined && template.group.id == userGroup.id
            );
        });
        if (templateGroupMatch == undefined) {
            //template not found -> all roles are invalid except the minimal role
            user.userGroups.push({ id: minimalGroupId, name: "Minimal Group" });
        }
        // this performs a **shallow** copy, in this case it should be ok. But sometimes you need a deep copy.
        return {...user};
    });
}
``` 

In case of `validateUsers` function, there is no return, so a forEach is in order.
The templateGroupMatch find can be a private method or a function to avoid repeating code.

README.md
```diff
@@ -486,15 +486,20 @@ List of excluded users (the script will ignore them) (could be an empty list [])
...
 A list with flags to control the script (permissionFixerConfig).
 
+A boolean variable push_report to determine if the script must send the report after processing all users
+
+A boolean variable pushFixedUserGroups to allow push the fixed usergroups (for those users without widp template user group)
+
+A boolean variable pushFixedUsersRoles to allow the push of the fixed users.
+
+A boolean variable forceMinimalGroupForUsersWithoutGroup to determine ignore the users with invalid usergroup exception, assuming for that users the minimal group
+
```
nshandra, line 502: Give this a grammar check (with GPT for example).

nshandra:
Also check `src/data/user-monitoring/permission-fixer/PermissionFixerReportD2Repository.ts` : 
- line 86 unnecessary await
- unused import

nshandra (changes requested):
Minor changes requiested.
Also check `src/data/user-monitoring/permission-fixer/PermissionFixerReportD2Repository.ts` : 
- line 86 unnecessary await
- unused import

src/data/user-monitoring/permission-fixer/PermissionFixerReportD2Repository.ts
```diff
@@ -71,27 +70,24 @@ export class PermissionFixerReportD2Repository implements PermissionFixerReportR
...
             JSON.stringify({ userToPost }, null, 4),
             filenameErrorOnPush
         );
         log.error(`Save errors in csv: `);
-        await UserMonitoringFileResourceUtils.savePermissionFixerToCsv(
-            userActionRequired,
-            `${csvErrorFilename}`
-        );
+        UserMonitoringFileResourceUtils.savePermissionFixerToCsv(userActionRequired, `${csvErrorFilename}`);
     }
```
nshandra, line 92:
```suggestion
        if (response?.status != "OK") {
            await this.saveUserErrorsOnLogFile(responseRoles.userProcessed);
            throw new Error("Error on push report: " + JSON.stringify(response));
        } else {
            return response.status;
        }
    }

    private async saveUserErrorsOnLogFile(userActionRequired: UserMonitoringUserResponse[]) {
        const userToPost: PermissionFixerUser[] = userActionRequired.map(item => {
            return item.fixedUser;
        });
        log.error(`Save jsons on import error: ${filenameErrorOnPush}`);
        UserMonitoringFileResourceUtils.saveInJsonFormat(
            JSON.stringify({ userToPost }, null, 4),
            filenameErrorOnPush
        );
        log.error(`Save errors in csv: `);
        await UserMonitoringFileResourceUtils.savePermissionFixerToCsv(
            userActionRequired,
            `${csvErrorFilename}`
        );
    }

```

`savePermissionFixerToCsv` is a async method, so its a good practice to await for it.

src/domain/usecases/user-monitoring/permission-fixer/RunUserPermissionUseCase.ts
```diff
@@ -102,7 +102,26 @@ export class RunUserPermissionUseCase {
...
+            const usersWithForcedMinimalGroup = this.addMinimalUserGroupToUsersWithoutUserGroup(
+                allUsers,
+                completeTemplateGroups,
+                minimalGroupId
+            );
+            return usersWithForcedMinimalGroup;
+        } else {
+            return allUsers;
+        }
+    }
```
nshandra, line 124:
```suggestion

    private preProcessUsers(
        allUsers: PermissionFixerUser[],
        completeTemplateGroups: PermissionFixerTemplateGroupExtended[],
        forceMinimalGroupForUsersWithoutGroup: boolean,
        minimalGroupId: string
    ) {
        if (forceMinimalGroupForUsersWithoutGroup) {
            log.info(
                "forceMinimalGroupForUsersWithoutGroup is enabled. Adding minimal group to users without group."
            );
            const usersWithForcedMinimalGroup = this.addMinimalUserGroupToUsersWithoutUserGroup(
                allUsers,
                completeTemplateGroups,
                minimalGroupId
            );
            return usersWithForcedMinimalGroup;
        } else {
            return allUsers;
        }
    }

```

Add spaces to separete the methods. Not sure why the code formatter didn't catch that.

PR #60, opened 2024-09-30
Feature/fix usergroup update

idelcano:
…usergroup

User permission fixer was adding to the WIDP USERS group all the users without group. But was overriding the userGroup sharing settings, also was using the endpoint that broke our instances.

I have changed from post the full userGroup to addition in the usergroup the missing users

 I tested it in a local instance using:
 yarn  start usermonitoring run-permissions-fixer --config-file config.json

### DIFF
--- a/src/data/user-monitoring/permission-fixer/PermissionFixerUserGroupD2Repository.ts
+++ b/src/data/user-monitoring/permission-fixer/PermissionFixerUserGroupD2Repository.ts
@@ -1,4 +1,4 @@
-import { D2Api, D2ApiResponse, Stats } from "@eyeseetea/d2-api/2.36";
+import { D2Api, Stats } from "@eyeseetea/d2-api/2.36";
 import log from "utils/log";
 import { PermissionFixerUserGroupExtended } from "domain/entities/user-monitoring/permission-fixer/PermissionFixerUserGroupExtended";
 import { PermissionFixerUserGroupRepository } from "domain/repositories/user-monitoring/permission-fixer/PermissionFixerUserGroupRepository";
--- a/src/data/user-monitoring/permission-fixer/PermissionFixerUserGroupD2Repository.ts
+++ b/src/data/user-monitoring/permission-fixer/PermissionFixerUserGroupD2Repository.ts
@@ -1,9 +1,10 @@
-import { D2Api } from "@eyeseetea/d2-api/2.36";
+import { D2Api, D2ApiResponse, Stats } from "@eyeseetea/d2-api/2.36";
 import log from "utils/log";
 import { PermissionFixerUserGroupExtended } from "domain/entities/user-monitoring/permission-fixer/PermissionFixerUserGroupExtended";
 import { PermissionFixerUserGroupRepository } from "domain/repositories/user-monitoring/permission-fixer/PermissionFixerUserGroupRepository";
 import { Async } from "domain/entities/Async";
 import { UserGroupNotFoundException } from "./exception/UserGroupNotFoundException";
+import { Id, Ref } from "domain/entities/Base";
 
 export class PermissionFixerUserGroupD2Repository implements PermissionFixerUserGroupRepository {
     constructor(private api: D2Api) {}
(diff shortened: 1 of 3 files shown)


src/data/user-monitoring/permission-fixer/PermissionFixerUserGroupD2Repository.ts
```diff
@@ -1,9 +1,10 @@
-import { D2Api } from "@eyeseetea/d2-api/2.36";
+import { D2Api, Stats } from "@eyeseetea/d2-api/2.36";
 import log from "utils/log";
 import { PermissionFixerUserGroupExtended } from "domain/entities/user-monitoring/permission-fixer/PermissionFixerUserGroupExtended";
 import { PermissionFixerUserGroupRepository } from "domain/repositories/user-monitoring/permission-fixer/PermissionFixerUserGroupRepository";
 import { Async } from "domain/entities/Async";
 import { UserGroupNotFoundException } from "./exception/UserGroupNotFoundException";
+import { Id, Ref } from "domain/entities/Base";
```
nshandra, line 7: `Id` is not used.

nshandra, line 7: Id is not used.

src/data/user-monitoring/permission-fixer/PermissionFixerUserGroupD2Repository.ts
```diff
@@ -23,21 +24,45 @@ export class PermissionFixerUserGroupD2Repository implements PermissionFixerUser
             throw new UserGroupNotFoundException("Error getting user group: " + groupsIds);
         }
     }
-    async save(userGroup: PermissionFixerUserGroupExtended): Async<string> {
+    async save(userGroup: PermissionFixerUserGroupExtended, users: Ref[]): Async<string> {
         try {
-            const response = await this.api.models.userGroups.put(userGroup).getData();
-            if (response.status == "OK") {
+            const response = await this.appendUsersInUsergroup(userGroup, users);
```
nshandra, line 29: Pedantic note: "in" usually means current location, "to" or "into" are more fitting.
So it should be `appendUsersToUsergroup` to accurately describe what the function does.

src/data/user-monitoring/permission-fixer/PermissionFixerUserGroupD2Repository.ts
```diff
@@ -23,21 +24,45 @@ export class PermissionFixerUserGroupD2Repository implements PermissionFixerUser
             throw new UserGroupNotFoundException("Error getting user group: " + groupsIds);
         }
     }
-    async save(userGroup: PermissionFixerUserGroupExtended): Async<string> {
+    async save(userGroup: PermissionFixerUserGroupExtended, users: Ref[]): Async<string> {
```
nshandra, line 27: You can leave it as it is, but this is not saving a userGroup, only adding members to it.

src/data/user-monitoring/permission-fixer/PermissionFixerUserGroupD2Repository.ts
```diff
@@ -23,21 +24,45 @@ export class PermissionFixerUserGroupD2Repository implements PermissionFixerUser
...
+
+    private async appendUsersInUsergroup(
+        userGroup: PermissionFixerUserGroupExtended,
+        users: Ref[]
+    ): Async<string> {
+        const usersIds = users.map(({ id }) => ({ id: id }));
+        const response = await this.api
+            .post<UserGroupResponse>(
+                `/userGroups/${userGroup.id}/users/`,
+                { async: false },
```
nshandra, line 53: Just a comment: as far as i know this does nothing as there is no async option for this endpoint.

nshandra (changes requested):
Just a couple of small comments, otherwise looks good.

src/data/user-monitoring/permission-fixer/PermissionFixerUserGroupD2Repository.ts
idelcano: na it's not  pedantic, thanks!

src/data/user-monitoring/permission-fixer/PermissionFixerUserGroupD2Repository.ts
idelcano: the reason to add tye async false is due without it I got a typo error

nshandra: I see, probably was developed with `api/metadata` in mind

PR #61, opened 2024-10-18
Feature/working tests

idelcano:
This pull request includes the necessary changes to run tests and the tests for the script that checks two-factor authentication.

Unfortunately, I can't get a working branch with the user-permissions test. 
I have two problems to finish that branch: 
- We changed some logic, and we need to add minimal tests to ensure nothing is broken. 
- When I execute the tests one by one, they pass. However, when I run all the tests together, three of them fail (the mocks return different objects compared to when I run them separately, and I see it at the end of the time).

The still in work branch:
https://github.com/EyeSeeTea/d2-tools/compare/development...feature/working_tests


xurxodev (approved):
Thanks @idelcano 

It's pending the second script for when we have time

xurxodev (approved):
thanks @idelcano 

I've only realised code review from the PR

PR #64, opened 2024-11-19
Feat: Close active Enrollments based on its events updatedAt date

nshandra:
### :pushpin: References

Issue: [[1035] Update enrollment status - OpenMRS implementation](https://app.clickup.com/t/8696mcy0h)

### :memo: Implementation

This script will get the events for a given orgUnit and Program, get the associated active enrollments and close them.

### DIFF
--- a/src/data/enrollments/EnrollmentsD2Repository.ts
+++ b/src/data/enrollments/EnrollmentsD2Repository.ts
@@ -26,6 +26,9 @@ export class EnrollmentsD2Repository implements EnrollmentsRepository {
                 ouMode: "SELECTED",
                 programStatus: "ACTIVE",
                 fields: enrollmentsRefFields,
+                // NOTE: Fix for 2.37.8.1
+                page: 1,
+                pageSize: 100000,
                 // TODO: Remove enrolledBefore once d2-api is updated
                 enrolledBefore: new Date().toISOString().replace("Z", ""),
                 skipPaging: true,
@@ -35,7 +38,7 @@ export class EnrollmentsD2Repository implements EnrollmentsRepository {
         return enrollments;
     }
 
-    // NOTE: This method will be usesd with bulkCloseEnrollments
+    // NOTE: This method will be used with bulkCloseEnrollments
     async getAllActive(params: EnrollmentsRepositoryParams): Async<Enrollment[]> {
         return this.getEnrollments({ ...params, programStatus: "ACTIVE" });
     }
@@ -55,6 +58,9 @@ export class EnrollmentsD2Repository implements EnrollmentsRepository {
                 ouMode: "SELECTED",
                 programStatus: programStatus,
                 fields: enrollmentsFields,
+                // NOTE: Fix for 2.37.8.1
+                page: 1,
+                pageSize: 100000,
                 // TODO: Remove enrolledBefore once d2-api is updated
                 enrolledBefore: new Date().toISOString().replace("Z", ""),
                 updatedAfter: updatedAfter,
--- a/src/data/enrollments/EnrollmentsD2Repository.ts
+++ b/src/data/enrollments/EnrollmentsD2Repository.ts
@@ -1,10 +1,11 @@
 import _ from "lodash";
 
 import { D2Api } from "types/d2-api";
+import { TrackerPostResponse } from "@eyeseetea/d2-api/api/tracker";
 
 import { Id } from "domain/entities/Base";
 import { Async } from "domain/entities/Async";
-import { Enrollment } from "domain/entities/enrollments/Enrollment";
+import { Enrollment, EnrollmentStatusRef, EnrollmentStatus } from "domain/entities/enrollments/Enrollment";
 
 import {
     EnrollmentsRepository,
@@ -15,37 +16,48 @@ import {
 export class EnrollmentsD2Repository implements EnrollmentsRepository {
     constructor(private api: D2Api) {}
 
-    async getList(params: EnrollmentsRepositoryParams, enrollmentsIds: Id[]): Async<Enrollment[]> {
+    async getAllActiveRef(params: EnrollmentsRepositoryParams): Async<EnrollmentStatusRef[]> {
         const { programId, orgUnitId } = params;
 
-        const idList = enrollmentsIds.join(";");
-
-        // NOTE: looks like @eyeseetea/d2-api/api/trackerEnrollments.d.ts: Params is not interpreted correctly and marked as required
         const { instances: enrollments } = (await this.api.tracker.enrollments
             .get({
                 program: programId,
                 orgUnit: orgUnitId,
                 ouMode: "SELECTED",
-                fields: { enrollment: true, status: true },
-                enrolledBefore: `${new Date().toISOString().replace("Z", "")}`,
-                enrollment: idList,
+                programStatus: "ACTIVE",
+                fields: enrollmentsRefFields,
+                // TODO: Remove enrolledBefore once d2-api is updated
+                enrolledBefore: new Date().toISOString().replace("Z", ""),
                 skipPaging: true,
             })
-            .getData()) as { instances: Enrollment[] };
+            .getData()) as { instances: EnrollmentStatusRef[] };
 
         return enrollments;
     }
 
-    async getAll(params: EnrollmentsRepositoryParams): Async<Enrollment[]> {
-        const { programId, orgUnitId } = params;
+    // NOTE: This method will be usesd with bulkCloseEnrollments
+    async getAllActive(params: EnrollmentsRepositoryParams): Async<Enrollment[]> {
+        return this.getEnrollments({ ...params, programStatus: "ACTIVE" });
+    }
+
+    async getRecentlyUpdated(params: EnrollmentsRepositoryParams): Async<Enrollment[]> {
+        const updatedAfter = getOneHourAgoDate();
+        return this.getEnrollments({ ...params, updatedAfter });
+    }
+
+    private async getEnrollments(params: getEnrollmentsParams): Async<Enrollment[]> {
+        const { programId, orgUnitId, programStatus, updatedAfter } = params;
 
         const { instances: enrollments } = (await this.api.tracker.enrollments
             .get({
                 program: programId,
                 orgUnit: orgUnitId,
                 ouMode: "SELECTED",
-                fields: { enrollment: true, status: true },
-                enrolledBefore: `${new Date().toISOString().replace("Z", "")}`,
+                programStatus: programStatus,
+                fields: enrollmentsFields,
+                // TODO: Remove enrolledBefore once d2-api is updated
+                enrolledBefore: new Date().toISOString().replace("Z", ""),
+                updatedAfter: updatedAfter,
                 skipPaging: true,
             })
             .getData()) as { instances: Enrollment[] };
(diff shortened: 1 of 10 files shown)


src/data/enrollments/EnrollmentsD2Repository.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export class EnrollmentsD2Repository implements EnrollmentsRepository {
+    constructor(private api: D2Api) {}
+
+    async getList(params: EnrollmentsRepositoryParams, enrollmentsIds: Id[]): Async<Enrollment[]> {
+        const { programId, orgUnitId } = params;
+
+        const idList = enrollmentsIds.join(";");
+
+        // NOTE: looks like @eyeseetea/d2-api/api/trackerEnrollments.d.ts: Params is not interpreted correctly and marked as required
```
tokland, line 23: I don't understand the note, which params is marked as required?

src/data/enrollments/EnrollmentsD2Repository.ts
```diff
@@ -0,0 +1,69 @@
...
+            .get({
+                program: programId,
+                orgUnit: orgUnitId,
+                ouMode: "SELECTED",
+                fields: { enrollment: true, status: true },
+                enrolledBefore: `${new Date().toISOString().replace("Z", "")}`,
+                enrollment: idList,
+                skipPaging: true,
+            })
+            .getData()) as { instances: Enrollment[] };
```
tokland, line 34: This will not be necessary when https://github.com/EyeSeeTea/d2-api/pull/151 is merged, as it infers the enrollments type from the fields. Let's keep as is.

src/data/enrollments/EnrollmentsD2Repository.ts
```diff
@@ -0,0 +1,69 @@
...
+        const idList = enrollmentsIds.join(";");
+
+        // NOTE: looks like @eyeseetea/d2-api/api/trackerEnrollments.d.ts: Params is not interpreted correctly and marked as required
+        const { instances: enrollments } = (await this.api.tracker.enrollments
+            .get({
+                program: programId,
+                orgUnit: orgUnitId,
+                ouMode: "SELECTED",
+                fields: { enrollment: true, status: true },
+                enrolledBefore: `${new Date().toISOString().replace("Z", "")}`,
```
tokland, line 30:
```
`${s}`
``` 
-> `s`

src/data/enrollments/EnrollmentsD2Repository.ts
```diff
@@ -0,0 +1,69 @@
...
+                skipPaging: true,
+            })
+            .getData()) as { instances: Enrollment[] };
+
+        return enrollments;
```
tokland, line 53: This code is exactly the same than getList except for the "enrollment" filter. We can abstract it (for example, creating a private method that's used by both)

src/data/enrollments/EnrollmentsD2Repository.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    async closeEnrollment(enrollmentId: Id): Async<TrackerResponse> {
+        const putResponse = await this.api.put<void>(`/enrollments/${enrollmentId}/completed`).response();
```
tokland, line 57: Not conceptually wrong, that PUT completed, by I wonder, is posting the enrollments with status: "COMPLETED" equivalent? that allows completing the N enrollments in one go, it will be much faster.

src/domain/entities/enrollments/Enrollment.ts
```diff
@@ -0,0 +1,8 @@
+import { Id } from "domain/entities/Base";
+
+export interface Enrollment {
+    enrollment: Id;
+    status: enrollmentStatus;
+}
+
+export type enrollmentStatus = "ACTIVE" | "COMPLETED" | "CANCELLED";
```
tokland, line 8: type casing: `EnrollmentStatus`

src/domain/repositories/enrollments/EnrollmentsRepository.ts
```diff
@@ -0,0 +1,20 @@
+import { Id } from "domain/entities/Base";
+import { Async } from "domain/entities/Async";
+import { Enrollment } from "domain/entities/enrollments/Enrollment";
+// import { Stats } from "domain/entities/Stats";
```
tokland, line 4: remove commented code

src/domain/repositories/enrollments/EventsRepository.ts
```diff
@@ -0,0 +1,14 @@
+import { Async } from "domain/entities/Async";
+import { Id } from "domain/entities/Base";
+// import { Stats } from "domain/entities/Stats";
```
tokland, line 3: same

src/scripts/common.ts
```diff
@@ -180,3 +180,18 @@ export const FilePath: Type<string, string> = {
...
+}
+
+export const MetadataDate: Type<string, string> = {
+    async from(str) {
+        if (!isValidDate(str)) throw new Error(`Invalid date format: ${str}`);
+        if (!str.includes("T")) {
+            str += "T00:00:00.000";
+        }
+        return str;
+    },
```
tokland, line 196: Not really important, but following best practices, that would look:

```ts
        if (!isValidDate(str)) {
            throw new Error(`Invalid date format: ${str}`);
        } else if (!str.includes("T")) {
            return str + "T00:00:00.000";
        } else {
            return str;
        }
```

src/scripts/commands/enrollments.ts
```diff
@@ -0,0 +1,52 @@
+import _ from "lodash";
+import { command, option, subcommands } from "cmd-ts";
+import { getApiUrlOption, getD2Api, IDString, MetadataDate } from "scripts/common";
+import { EventsD2Repository } from "data/enrollments/EventsD2Repository";
+// import { GetEnrollmentsIDsToCloseUseCase } from "domain/usecases/enrollments/GetEnrollmentsIDsToCloseUseCase";
```
tokland, line 5: remove

src/scripts/commands/enrollments.ts
```diff
@@ -0,0 +1,52 @@
...
+        }),
+        programs: option({
+            type: IDString,
+            long: "program-id",
+            description: "Program(s) of the enrollments, comma-separated",
+        }),
+        eventUpdateCutoff: option({
+            type: MetadataDate,
+            long: "event-date-before",
+            description: "YYYY-MM-DDThh:mm:ss",
```
tokland, line 37: Hour is optional, right? typically it's written like this: `"YYYY-MM-DD[Thh:mm:ss]"`

src/scripts/commands/enrollments.ts
```diff
@@ -0,0 +1,52 @@
...
+        url: getApiUrlOption({ long: "url" }),
+        orgUnits: option({
+            type: IDString,
+            long: "org-unit-id",
+            description: "Organization Unit(s) of the enrollments, comma-separated",
+        }),
+        programs: option({
+            type: IDString,
+            long: "program-id",
+            description: "Program(s) of the enrollments, comma-separated",
```
tokland, line 32: both are single values, not comma-separated. And then the name is singular (orgUnitId, programId)

src/domain/usecases/enrollments/CloseEnrollmentsUseCase.ts
```diff
@@ -0,0 +1,59 @@
...
+
+        if (activeEnrollmentsIDs.length === 0) {
+            log.info("No active Enrollments found for the provided date");
+            return;
+        }
+
+        log.info(`Found ${activeEnrollmentsIDs.length} active enrollments to close`);
+
+        activeEnrollmentsIDs.forEach(async enrollmentId => {
+            const report: TrackerResponse = await this.enrollmentsRepository.closeEnrollment(enrollmentId);
```
tokland, line 49: That's a typical problem, forEach+await does not work as expected. ChatGPT gives a nice explanation as for why. We have the abstraction promiseMap for this scenario.

If we can finally post the enrollments in a single payload we won't need this, though.

src/domain/usecases/enrollments/CloseEnrollmentsUseCase.ts
```diff
@@ -0,0 +1,59 @@
...
+
+            if (report.status === "ERROR") {
+                console.error(report.message);
+            }
+
+            // NOTE: Added to not overload the server
+            await new Promise(resolve => setTimeout(resolve, 25));
```
tokland, line 56: If posting all the enrollments in one payload, we won't need this

tokland (commented):
In-line comments:

src/data/enrollments/EnrollmentsD2Repository.ts
nshandra: image
It make sense that its something on my end because the type definition in node_modules/@eyeseetea/d2-api/api/trackerEnrollments.d.ts looks OK, but still, i get this error.

src/data/enrollments/EnrollmentsD2Repository.ts
nshandra: ....facepalm moment. Fixed.

src/scripts/commands/enrollments.ts
nshandra: Right.

src/domain/usecases/enrollments/CloseEnrollmentsUseCase.ts
nshandra: I remember now that this was touched upon in a dojo (i think). Anyway, we could use a promiseMap in 2.37 for performance reasons, [see](https://github.com/EyeSeeTea/d2-tools/pull/64#discussion_r1850949673). Maybe control the enrollment update method via an option.

src/domain/usecases/enrollments/CloseEnrollmentsUseCase.ts
nshandra: I guess using the promiseMap is not needed either really.

src/data/enrollments/EnrollmentsD2Repository.ts
tokland: Oh, I see, that's indeed a bug in d2-api, and hard to track down because of the cryptic error message (the source of the problem: (TrackerEnrollmentsParamsBase["orgUnit"] & PartialParams)). That should be a pick. This has been fixed in the same PR I referenced before, so nothing to worry about.

src/data/enrollments/EnrollmentsD2Repository.ts
tokland: Ok, let's go with the PUTs then

src/domain/usecases/enrollments/CloseEnrollmentsUseCase.ts
tokland: Yes, I guess the error came from the hundred of parallel requests (as forEach does not await).

tokland:
@nshandra I guess you're still with the changes (no new commits), click re-review when ready

src/data/enrollments/EnrollmentsD2Repository.ts
nshandra: The repository was updated to be able to switch to use PUT or POST closing methods. Ideally one does not add methods that are not used, but in this case i already had them developed to test the performance.

tokland (approved):
Cool. I've just remove the prop stats from the domain entity, which is not used (and it was any type). Whenever we need it, we'd re-add with specific type (it was "any")

PR #67, opened 2025-01-05
[new script] [programs] Move data values from old data elements to new data elements on specific tracker program's program stage

p3rcypj:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/86975nz4f

### :memo: Implementation

Added script in order to move data values from certain program stage without modyifing old values. It should just add new data values with same properties but with the data element id as the target data element specified (from the pair specified)

Source_DE1 ➝ Target_DE3
Source_DE2 ➝ Target_DE4

### :fire: Testing

```console
yarn start:dev programs move-data-values \
    --url=http://$DHIS2_USER:$DHIS2_PASSWORD@localhost:8080 \
    --program-stage-id=sgSKZRoWE9b \
    --save-report='./move-data-values-report.txt' \
    qwfxR2TQkUn-L5x9z9BAgR8 \
    N0p2yOsEy7a-wSCh46cADN6 \
    R1MphCO50la-DrTaMDJAQoS \
    FHBb7rMzYEL-s8hgq0Y8SJs
```
PS.: add `--post` to send the payload

Docker: `docker.eyeseetea.com/ocg/dhis2-data:2.37.8-pro`

Added tests using Copilot Edits:
- should move data values successfully
- should throw error if data element types do not match
- should throw error if some data elements are missing
- should throw error if target data values are not empty
- should return empty array if there is no data value with some source data element id


p3rcypj:
Not sure on which section should I add an example of this script on the README

src/data/DataElementsD2Repository.ts
```diff
@@ -1,23 +1,31 @@
...
+import { DataElement } from "domain/entities/DataElement";
 
 export class DataElementsD2Repository implements DataElementsRepository {
     constructor(private api: D2Api) {}
 
+    async getByIds(ids: Id[]): Promise<DataElement[]> {
+        return this.getDataElements(ids);
+    }
+
     async getDataElementsNames(ids: Id[]): Promise<NamedRef[]> {
```
tokland, line 14: `getByIds` is returning `DataElement[]`, which makes a lot more sense that `getDataElementsNames` returning `NamedRef[]`. And, since DataElement is compatible with NamedRef, I'd remove getDataElementsNames and make everybody use getByIds.

src/data/DataElementsD2Repository.ts
```diff
@@ -1,23 +1,31 @@
...
-                    id: true,
-                    name: true,
-                },
+                fields: dataElementFields,
                 filter: { id: { in: ids } },
             },
         });
 
-        const { dataElements } = await metadata$.getData();
+        const dataElements = (await metadata$.getData()).dataElements;
```
tokland, line 28: why that change? the original looks more declarative to me

src/domain/usecases/CopyProgramStageDataValuesUseCase.ts
```diff
@@ -0,0 +1,175 @@
...
+            fs.writeFileSync(payloadPath, json);
+            log.info(`Written payload (${eventsWithNewDataValues.length} events): ${payloadPath}`);
+        }
+
+        if (reportPath) {
+            this.saveReport(reportPath, dataElementPairs, programStageId, eventsWithNewDataValues);
+        }
+
+        return eventsWithNewDataValues;
+    }
```
tokland, line 66: That's too long, specially being the public method of the class. Identify the steps involved and abstract to private methods.

src/domain/usecases/CopyProgramStageDataValuesUseCase.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    private copyEventDataValues(
+        applicableEvents: ProgramEvent[],
+        sourceIds: string[],
+        dataElementPairs: DataElementPair[]
+    ) {
```
tokland, line 72: As a rule of thumb, put always the explicit return type.

src/domain/usecases/CopyProgramStageDataValuesUseCase.ts
```diff
@@ -0,0 +1,175 @@
...
+                if (!target)
+                    throw new Error(`Target data element not found for source id: ${dv.dataElement.id}`);
+
+                return [
+                    dv,
+                    {
+                        ...dv,
+                        dataElement: _.omit(target, "valueType"),
+                    },
+                ];
```
tokland, line 88: The layout of conditional blocks is one of the most important things to be declarative. If you are able to put a single if/else if/else chain, it will look nicer. Proposal: 

```
                const target = dataElementPairs.find(([source, _]) => source.id === dv.dataElement.id)?.[1];

                if (!sourceIds.includes(dv.dataElement.id)) {
                    return [dv];
                } else if (!target) {
                    throw new Error(`Target data element not found for source id: ${dv.dataElement.id}`);
                } else {
                    return [dv, { ...dv, dataElement: _.omit(target, "valueType") }];
                }
```

src/domain/usecases/CopyProgramStageDataValuesUseCase.ts
```diff
@@ -0,0 +1,175 @@
...
+        .compact()
+        .join("\n");
+
+    const error = `Some data values of the destination data elements are not empty:\n${eventsWithNonEmptyTargetDataValues}`;
+    if (eventsWithNonEmptyTargetDataValues) throw new Error(error);
+}
+
+export type CopyProgramStageDataValuesOptions = {
+    programStageId: string;
+    dataElementIdPairs: [Id, Id][]; // [sourceDataElementId, targetDataElementId]
```
tokland, line 170: Working with tuples should be avoided, all the code interacting with them will be uglier. And you see, you had to add a comment about what it holds. So the most declarative structure is an object: `type DataElementMapping = {sourceId: Id, targetId: Id}`

src/domain/usecases/CopyProgramStageDataValuesUseCase.ts
```diff
@@ -0,0 +1,175 @@
...
+    post: boolean;
+    saveReport?: string;
+};
+
+type DataElementPair = [DataElement, DataElement];
```
tokland, line 175: And the same for this type.

src/scripts/commands/programs.ts
```diff
@@ -231,6 +244,47 @@ const moveAttribute = command({
...
+            description: "Save JSON payload",
+        }),
+    },
+    handler: async args => {
+        const api = getD2Api(args.url);
+        const programEventsRepository = new ProgramEventsD2Repository(api);
+        const dataElementsRepository = new DataElementsD2Repository(api);
+        const orgUnitRepository = new OrgUnitD2Repository(api);
+
+        new CopyProgramStageDataValuesUseCase(
```
tokland, line 280: await

tokland (commented):
In-line comments:

src/data/DataElementsD2Repository.ts
p3rcypj: It is indeed to me also. I made the change while testing because it was more comfortable for me but I forgot to reverse it

src/domain/usecases/CopyProgramStageDataValuesUseCase.ts
```diff
@@ -167,9 +179,9 @@ function checkTargetDataValuesAreEmpty(events: ProgramEvent[], targetIds: Id[])
 
 export type CopyProgramStageDataValuesOptions = {
     programStageId: string;
-    dataElementIdPairs: [Id, Id][]; // [sourceDataElementId, targetDataElementId]
+    dataElementIdMappings: { source: Id; target: Id }[];
```
tokland, line 182: don't change, but in that case where we have an object we typically use Array<...>

tokland (approved):
All good, code-wise!

PR #68, opened 2025-01-08
[Internal] Add CodeQL verification

MatiasArriola:
Added CodeQL Github Action


ifoche (approved):
thanks @MatiasArriola . Tests are broken but they're not broken by this PR so approving it

PR #77, opened 2025-04-14
DataSets: Set skip offline based on last input year

gqcorneby:
Required by: https://app.clickup.com/t/8698nvdmu

**Set skip offline**
- add command and script to set dataset skip offline

**NOTE:**
- had to include opening and closing data per dataInputPeriod to prevent them from being cleared when posting the data
- These properties were getting cleared by `d2-tools` because they aren’t included in the fields being fetched. When the data is posted, since these properties are missing from the payload, they get overwritten (cleared).
- This same clearing behavior also happened previously (before `set-skip-offline` was introduced) when running the `copy-org-units` command — specifically for the destination dataset.

### DIFF
--- a/src/domain/usecases/SetSkipOfflineDataSetUseCase.ts
+++ b/src/domain/usecases/SetSkipOfflineDataSetUseCase.ts
@@ -1,19 +1,19 @@
 import _ from "lodash";
 
-import { DataSetsRepository } from "../repositories/DataSetsRepository";
-import { Maybe } from "../../utils/ts-utils";
+import { DataSetPostResult, DataSetsRepository } from "../repositories/DataSetsRepository";
 import { DataSet } from "../entities/DataSet";
 import log from "../../utils/log";
 
 type SetSkipOfflineOptions = {
     year: number;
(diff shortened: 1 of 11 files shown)


src/domain/usecases/SetSkipOfflineDataSetUseCase.ts
```diff
@@ -0,0 +1,55 @@
...
+type SetSkipOfflineOptions = {
+    year: number;
+    disable: Maybe<boolean>;
+};
+
+export class SetSkipOfflineDataSetUseCase {
+    constructor(private dataSetRepository: DataSetsRepository) {}
+
+    async execute(options: SetSkipOfflineOptions) {
+        const updatedDataSets = await this.dataSetRepository
```
eperedo, line 17: Adding a log like 'Loading dataSets...' here could be useful, since fetching all the dataSets from the server might take a long time.

eperedo (changes requested):
Thanks @gqcorneby, looks good! Just a minor comment about clean arch. and a few typescript improvements.

PR #80, opened 2025-05-02
add analyze mode for options

eperedo:
### :pushpin: References

- **Issue:** Closes https://app.clickup.com/t/8698pp7e5

### :memo: Implementation

- [x] add `analyze` mode for rename option script.

For codes in option:

- [x] validate length (230 by default, you can customize the value using the `----code-name-length` flag.
- [x] check duplicate codes (same code in different option sets)
- [x] validate if code/name includes spaces, special misplaced commas or incorrect order.

**UPDATE 2025-05-14: Added validation for option names**

```bash
yarn start options analyze --url='http://localhost:8080' --auth='admin:district'
```

Csv report will be generated at `option-report.csv`. You can use the `--report-path` to save the report in a different path

#8698pp7e5


src/domain/entities/OptionValidationResult.ts
```diff
@@ -0,0 +1,15 @@
...
+
+type RuleType = "duplicate" | "naming_conventions" | "invalid_length" | "misplaced_commas" | "order";
+
+export type OptionValidationResult = {
+    option: Option;
+    optionSet: Omit<OptionSet, "options">;
+    rules: OptionValidationRule[];
+};
+
+export type OptionValidationRule = {
```
tokland, line 12: a "rule" or "validationRule" conveys the idea of definition, but here we are describing only problems, right? let's add that to prop "rules" and name "OptionValidationRule" (error? unmet? invalid?)

src/data/OptionSetD2Repository.ts
```diff
@@ -0,0 +1,44 @@
...
+    private getOptionsSets(page: number) {
+        return this.api.models.optionSets
+            .get({ fields: optionSetFields, pageSize: PAGE_SIZE, page: page })
+            .getData();
+    }
+
+    private async getTotalPages(options: { pageSize: number }): Async<number> {
+        const response = await this.api.models.optionSets
+            .get({ fields: { id: true }, pageSize: options.pageSize })
+            .getData();
```
tokland, line 32: Not that it's a problem, but it's a pity to throw away information we requested. I'd set pageSize=0 in the request and calculate the number of pages from prop "total", this would work, right?

src/domain/usecases/ValidateOptionSetsUseCase.ts
```diff
@@ -0,0 +1,106 @@
...
+import { OptionSetRepository } from "domain/repositories/OptionSetRepository";
+import logger from "utils/log";
+import { Async } from "domain/entities/Async";
+import { Option } from "domain/entities/Option";
+
+export class ValidateOptionSetsUseCase {
+    constructor(private readonly optionSetRepository: OptionSetRepository) {}
+
+    async execute(options: UseCaseOptions): Async<OptionValidationResult[]> {
+        logger.info("Loading optionSets...");
```
tokland, line 13: Let's move the logger.info to the method doing the action. The cleaner we keep the public method, the better.

src/domain/usecases/ValidateOptionSetsUseCase.ts
```diff
@@ -0,0 +1,106 @@
...
+                    ...orderRules,
+                ],
+            };
+        });
+    }
+
+    private validateLength(option: Option, validationLength: number): OptionValidationRule[] {
+        return option.code.length > validationLength
+            ? [{ type: "invalid_length", message: `Code length is ${option.code.length}` }]
+            : [];
```
tokland, line 56: This check conceptually will never be returning N>1 validations, right? then the signature should reflect it: `Maybe<OptionValidationRule>`

src/domain/usecases/ValidateOptionSetsUseCase.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    private validateNamingConventions(option: Option): OptionValidationRule[] {
+        const hasSpaces = /\s/.test(option.code);
+        const hasSpecialCharacters = option.code.match(/[^\w\s]/) ?? [];
```
tokland, line 61: it's an array, so probably it should not have a "has" prefix.

src/domain/usecases/ValidateOptionSetsUseCase.ts
```diff
@@ -0,0 +1,106 @@
...
+
+        const spacesRule = hasSpaces ? "spaces" : "";
+        const specialRule =
+            hasSpecialCharacters.length > 0 ? `special characters: ${hasSpecialCharacters.join(",")}` : "";
```
tokland, line 65: We typically use "undefined" to convey "no value", instead of empty string.

src/domain/usecases/ValidateOptionSetsUseCase.ts
```diff
@@ -0,0 +1,106 @@
...
+        const message = [spacesRule, specialRule].filter(Boolean).join(" and ");
+
+        return spacesRule || specialRule
+            ? [{ type: "naming_conventions", message: `Code contains ${message}` }]
+            : [];
+    }
+
+    private validateMisplacedCommas(option: Option): OptionValidationRule[] {
+        return option.code.includes(",")
+            ? [{ type: "misplaced_commas", message: `Code has misplaced commas` }]
```
tokland, line 76: maybe it comes from the spec, but it looks weird, the text is "misplaced commas", but the code condition will be true for any comma found in the code, regardless of its place.

src/domain/usecases/ValidateOptionSetsUseCase.ts
```diff
@@ -0,0 +1,106 @@
...
+        return option.sortOrder !== expected
+            ? [{ type: "order", message: `sortOrder is ${option.sortOrder}; expected ${expected}` }]
+            : [];
+    }
+
+    private getDuplicateOptions(optionSets: OptionSet[]): OptionValidationResult[] {
+        return _(optionSets)
+            .flatMap(optionSet => optionSet.options.map(option => ({ option: option, optionSet: optionSet })))
+            .groupBy(item => item.option.code)
+            .pickBy(group => group.length > 1)
```
tokland, line 91: [subjective] when we don't need the group key, the more declarative processing is: .values()  + .filter()

src/utils/options.ts
```diff
@@ -0,0 +1,46 @@
+import { OptionValidationResult } from "domain/entities/OptionValidationResult";
+import _ from "lodash";
+
+function escapeCsv(value: string): string {
+    return `"${value.replace(/"/g, '""')}"`;
+}
```
tokland, line 6: CSV is a simple format, but still, a complete implementation has to take in account quotes, multi-lines and so on. I'd just use a csv library.

src/utils/options.ts
```diff
@@ -0,0 +1,46 @@
+import { OptionValidationResult } from "domain/entities/OptionValidationResult";
```
tokland, line 1: Let's try to avoid a src/utils file. This is being used only from src/scripts/commands. I'd put the file there (in some folder?) and a filename as declarative as possible.  And the function/class can also call writeFileSync.

src/utils/options.ts
```diff
@@ -0,0 +1,46 @@
...
+                optionSet.name,
+                optionSet.code ?? "",
+                optionSet.id,
+            ]
+                .map(escapeCsv)
+                .join(",");
+
+            const spaceBetweenOptionAndRules = "";
+
+            const ruleHeaders = ["", "rule type", "rule message"].map(escapeCsv).join(",");
```
tokland, line 35: We end with 2 consecutive spaces here?

src/scripts/commands/options.ts
```diff
@@ -1,22 +1,64 @@
...
+        }
+
+        if (args.optionId && args.toCode) {
+            const optionRepository = new OptionD2Repository(api);
+            await new RenameOptionCodeUseCase(optionRepository).execute(args);
+        } else {
+            logger.info(
+                `Provide a valid option ID and a new code to rename the option: --id <optionId> --to-code <newCode>`
+            );
+        }
```
tokland, line 61: To me, this look like two different sub-commands, one for rename, the other for analyze, right? This way we will also be able to enforce better which options are required.

src/scripts/commands/options.ts
```diff
@@ -1,22 +1,64 @@
...
+            defaultValue: () => "option-report.csv",
+        }),
+        codeLength: option({
+            type: number,
+            long: "code-length",
+            description: "Length of the code to validate",
+            defaultValue: () => 230,
+        }),
     },
     handler: async args => {
```
tokland, line 38: I see that we are putting all commands in the same file. I think it would be more readable to create for example a src/scripts/commands/options and create a file for each command there.

tokland (commented):
In-line comments.

Check with Adrian if we time for adding tests, I think that this kind of use cases, with a lot of logic, would benefit from having some specs.

src/domain/entities/Option.ts
```diff
@@ -6,3 +6,5 @@ export type Option = {
     code: string;
     sortOrder: number;
 };
+
+export const DEFAULT_VALID_LENGTH = 230;
```
tokland, line 10: [nothing to do] That's the limit imposed by DHIS2, but, as we are building d2 tools, we can say it's ok to impose it  in the model.

src/scripts/options/rename.ts
```diff
@@ -0,0 +1,61 @@
...
+import { OptionD2Repository } from "data/OptionD2Repository";
+import { RenameOptionCodeUseCase } from "domain/usecases/RenameOptionCodeUseCase";
+import { ValidateOptionSetsUseCase } from "domain/usecases/ValidateOptionSetsUseCase";
+import { OptionSetD2Repository } from "data/OptionSetD2Repository";
+import { generateOptionReport } from "./csv";
+import logger from "utils/log";
+import { DEFAULT_VALID_LENGTH } from "domain/entities/Option";
+
+export const renameCodeCmd = command({
+    name: "rename",
```
tokland, line 13: I'm probably missing the big picture, but about my previous comment:

https://github.com/EyeSeeTea/d2-tools/pull/80#discussion_r2071373686

What's the idea?

src/scripts/options/csv.ts
```diff
@@ -0,0 +1,41 @@
+import fs from "fs/promises";
```
tokland, line 1: Let's be more precise with the name. As we use a free function, the only thing exported: generateOptionReport.ts. If using a class: OptionReport.ts (which would have a single public method: optionReport.generate())

src/domain/usecases/ValidateOptionSetsUseCase.ts
```diff
@@ -37,66 +36,69 @@ export class ValidateOptionSetsUseCase {
...
+            ? { type: "naming_conventions", message: `Code contains ${message}` }
+            : undefined;
     }
 
-    private validateMisplacedCommas(option: Option): OptionValidationRule[] {
-        return option.code.includes(",")
-            ? [{ type: "misplaced_commas", message: `Code has misplaced commas` }]
-            : [];
+    private validateMisplacedCommas(option: Option): Maybe<OptionValidationError> {
+        const hasMisplacedCommas = /^,|,,|,$|,/.test(option.code);
```
tokland, line 75: Regexp are not declarative enough, so we'd need a comment of what it's considered a misplaced comma. 

It's weird, because we have at-start, double, at-end, but also ",", which is any comma, right? so the previous checks are not useful.

tokland (commented):
In-line comments:

src/scripts/options/rename.ts
eperedo: Moved into its own sub-command.

README.md
```diff
@@ -936,6 +936,23 @@ yarn start options rename-code \
...
+
+Analyze option codes and generates a csv report if codes:
+
+-   Are duplicated in other optionSets
+-   Include any special characters (only underscore is allowed)
+-   length is greater than 230 (you can change this value --code-length [VALUE])
+-   have spaces or commas
+
+```shell
+yarn start options analyze-codes \
```
tokland, line 949: Ok for now, if we end up checking more codes, we can rename it for something more general like "validate"

src/domain/usecases/ValidateOptionSetsUseCase.ts
```diff
@@ -71,10 +71,9 @@ export class ValidateOptionSetsUseCase {
             : undefined;
     }
 
-    private validateMisplacedCommas(option: Option): Maybe<OptionValidationError> {
-        const hasMisplacedCommas = /^,|,,|,$|,/.test(option.code);
-        return hasMisplacedCommas
-            ? { type: "misplaced_commas", message: "Code has misplaced commas" }
+    private validateCommas(option: Option): Maybe<OptionValidationError> {
+        return option.code.includes(",")
+            ? { type: "misplaced_commas", message: "Code has commas" }
```
tokland, line 76: maybe we can rename also the type

src/scripts/options/analyze.ts
```diff
@@ -0,0 +1,42 @@
...
+    handler: async args => {
+        const api = getD2ApiFromArgs(args);
+        const optionSetRepository = new OptionSetD2Repository(api);
+
+        const validationResult = await new ValidateOptionSetsUseCase(optionSetRepository).execute({
+            codeLength: args.codeLength,
+        });
+
+        if (validationResult.length > 0) {
+            await new OptionReport().generateReport(validationResult, args.reportPath);
```
tokland, line 37: Maybe we can exit here with a 1 to signal error.

tokland (approved):
Extremely minor things, so approving, thx.

README.md
```diff
@@ -946,11 +946,11 @@ Analyze option codes and generates a csv report if codes:
 -   have spaces or commas
 
 ```shell
-yarn start options analyze-codes \
+yarn start options analyze \
     --url='http://localhost:8080' \
     --auth='username:password' \
     --report-path='report-name.csv' \  # default is option-report.csv
-    --code-length=20 # default is 230
+    --length=20 # default is 230
```
tokland: `--length` is too generic, which length should the user expect to be validated? It's name/code, so not sure if put --code-name-length or what, tbh.

src/data/D2RenameOptionCode.ts
```diff
@@ -231,6 +269,27 @@ export class D2RenameOptionCode {
...
     }
+
+    private async rollback(options: OptionsWithMetadataValues) {
+        const { option, metadata, initialDataValues, initialEvents } = options;
+
+        console.debug("[rollback] Executing rollback...");
+
+        await this.saveOption({ ...options, toCode: option.code, metadata });
+        await this.recodeDataValuesPost(initialDataValues);
+        await this.recodeEventsPost(initialEvents);
```
tokland: Now that these methods are not specifically recoding, but just posting, let's rename them? maybe postDataValues, postEvents

src/data/D2RenameOptionCode.ts
```diff
@@ -231,6 +269,27 @@ export class D2RenameOptionCode {
...
+
+        console.debug("[rollback] Rollback completed");
+    }
+
+    private saveDataToDisk(options: OptionsWithMetadataValues): void {
+        const { option, initialDataValues, initialEvents } = options;
+
+        saveJsonToDisk(`dataValues_${option.id}`, { dataValues: initialDataValues });
+        saveJsonToDisk(`events_${option.id}.json`, { events: initialEvents });
```
tokland: the json extension is not necessary, right?

src/data/D2RenameOptionCode.ts
```diff
@@ -15,6 +15,7 @@ import { D2TrackerEventToPost } from "@eyeseetea/d2-api/api/trackerEvents";
  * 1) Retrieve the data
  * 2) Update the option metadata
  * 3) Update the data
+ * 4) rollback if an error occurs (save json to disk and persist initial data)
```
tokland, line 18: The idea is to persist the data in case the rollback also fails, right? we can put the idea in the comment, so we know the files are not really used in the script, it's just a backup.

src/scripts/options/analyze.ts
```diff
@@ -18,23 +19,32 @@ export const analyzeCodesCmd = command({
...
+            long: "length",
+            description: "Max length of the code/name for it to be considered valid",
             defaultValue: () => DEFAULT_VALID_LENGTH,
         }),
+        update: flag({
+            long: "update",
+            description: "Persist changes",
+            defaultValue: () => false,
+            type: boolean,
+        }),
```
tokland, line 33: Is this update option used?

src/data/D2RenameOptionCode.ts
```diff
@@ -29,12 +30,35 @@ export class D2RenameOptionCode {
...
+
+        this.saveDataToDisk(optionsWithMetadataValues);
+
+        await this.recodeOption(optionsWithMetadataValues).catch(err => {
+            console.debug(`Error recoding option: ${JSON.stringify(err, null, 4)}`);
+            return this.rollback(optionsWithMetadataValues);
+        });
+    }
+
+    private async getValuesForOption(options: RecodeOptionsWithMetadata): Async<OptionsWithMetadataValues> {
```
tokland: `getValuesForOption`  sounds weird for a method that returns a `OptionsWithMetadataValues`

src/domain/usecases/ValidateOptionSetsUseCase.ts
```diff
@@ -31,34 +64,69 @@ export class ValidateOptionSetsUseCase {
...
+                return { option, optionSet, errors: allOptionsErrors };
+            });
+
+            return _(allPropertiesValidationResult)
+                .groupBy(x => x.option.id)
+                .map((group): OptionValidationResult => {
+                    const option = group[0]?.option;
+                    const optionSet = group[0]?.optionSet;
+                    if (!optionSet) throw new Error("No optionSet found");
+                    if (!option) throw new Error("No option found");
```
tokland: the typical approach is set group[0] to a const, and guard on that. Code will be shorter and more declarative

tokland (commented):
In-line:

tokland (approved):

PR #86, opened 2025-06-18
Fix twofactor user groups report Feature/add twofactor new options

idelcano:
Required:
https://github.com/EyeSeeTea/d2-api/pull/171

The d2-api is updated and this branch could be merged, so this PR is ready to be taken out of draft.

It includes updated logic and tests to meet the new requirements:

- Detect users who have not enabled two-factor authentication.

- Disable those users if the --disable-users CLI flag is set.

It also expands the reporting to cover all authentication-related issues, including:

- WHO users without WHO external access enabled.

- Users belonging to both WHO and Two-Factor user groups.

- Any other invalid group combinations.

Previously, the report only listed users in the Two-Factor user group, which was not sufficient for the final workflow.

The functional test for approving this PR may not be necessary, as the code is already deployed to production and working as expected.

### DIFF
--- a/src/data/user-monitoring/two-factor-monitoring/TwoFactorReportD2Repository.ts
+++ b/src/data/user-monitoring/two-factor-monitoring/TwoFactorReportD2Repository.ts
@@ -21,31 +21,23 @@ type ServerResponse = { status: string; typeReports: object[] };
 export class TwoFactorReportD2Repository implements TwoFactorReportRepository {
     constructor(private api: D2Api) {}
     async save(program: UserMonitoringProgramMetadata, report: TwoFactorUserReport): Async<string> {
+        const invalidTwoFAList = this.formatUsers(report.invalidTwoFAList);
+        const invalidWhoList = this.formatUsers(report.invalidWhoList);
+        const invalidAuthList = this.formatUsers(report.invalidAuthList);
+        
         const twoFactorUsersFileResourceId = await UserMonitoringFileResourceUtils.saveFileResource(
-            report.invalidTwoFAList
-                .map(user => {
-                    return user.name + "," + user.id;
-                })
-                .join("\n"),
+            invalidTwoFAList,
             "_twoFa"+filenameUserReported,
             this.api
         );
 
         const whoAccountUsersFileResourceId = await UserMonitoringFileResourceUtils.saveFileResource(
-            report.invalidWhoList
-                .map(user => {
-                    return user.name + "," + user.id;
-                })
-                .join("\n"),
+            invalidWhoList,
             "_who"+filenameUserReported,
             this.api
         );
         const invalidUsersFileResourceId = await UserMonitoringFileResourceUtils.saveFileResource(
-            report.invalidAuthList
-                .map(user => {
-                    return user.name + "," + user.id;
-                })
-                .join("\n"),
+            invalidAuthList,
             "_invalid"+filenameUserReported,
             this.api
         );
@@ -66,7 +58,13 @@ export class TwoFactorReportD2Repository implements TwoFactorReportRepository {
             return response.status;
         }
     }
-
+    
+    private formatUsers(users: { name: string; id: string }[]): string {
+        return users
+            .map(user => `${user.name},${user.id}`)
+            .join("\n");
+    }
+    
     private async push(
         invalidConfigNumber: string,
         invalidFileReferenceListUsers: string,
--- a/src/data/user-monitoring/two-factor-monitoring/TwoFactorReportD2Repository.ts
+++ b/src/data/user-monitoring/two-factor-monitoring/TwoFactorReportD2Repository.ts
@@ -10,6 +10,11 @@ import { UserMonitoringFileResourceUtils } from "../common/UserMonitoringFileRes
 
 const dataelement_invalid_two_factor_count_code = "ADMIN_users_without_two_factor_count_7_Events";
 const dataelement_invalid_two_factor_usernames_list_code = "ADMIN_users_without_two_factor_8_Events";
+const dataelement_invalid_who_accounts_list = "ADMIN_users_with_invalid_who_account_9_Events";
+const dataelement_invalid_who_accounts = "ADMIN_users_with_invalid_who_account_count_10";
+const dataelement_invalid_auth_list = "ADMIN_users_without_auth_groups_11_Events";
+const dataelement_invalid_auth = "ADMIN_users_without_auth_groups_count_12_Events";
+
 const filenameUserReported = `_users_reported.csv`;
 type ServerResponse = { status: string; typeReports: object[] };
 
@@ -17,17 +22,40 @@ export class TwoFactorReportD2Repository implements TwoFactorReportRepository {
     constructor(private api: D2Api) {}
     async save(program: UserMonitoringProgramMetadata, report: TwoFactorUserReport): Async<string> {
         const twoFactorUsersFileResourceId = await UserMonitoringFileResourceUtils.saveFileResource(
-            report.listOfAffectedUsers
+            report.invalidTwoFAList
                 .map(user => {
                     return user.name + "," + user.id;
                 })
                 .join("\n"),
-            filenameUserReported,
+            "_twoFa"+filenameUserReported,
+            this.api
+        );
+
+        const whoAccountUsersFileResourceId = await UserMonitoringFileResourceUtils.saveFileResource(
+            report.invalidWhoList
+                .map(user => {
+                    return user.name + "," + user.id;
+                })
+                .join("\n"),
+            "_who"+filenameUserReported,
+            this.api
+        );
+        const invalidUsersFileResourceId = await UserMonitoringFileResourceUtils.saveFileResource(
+            report.invalidAuthList
+                .map(user => {
+                    return user.name + "," + user.id;
+                })
+                .join("\n"),
+            "_invalid"+filenameUserReported,
             this.api
         );
         const response = await this.push(
-            report.invalidUsersCount.toString(),
+            report.invalidTwoFACount.toString(),
             twoFactorUsersFileResourceId,
+            report.invalidWhoCount.toString(),
+            whoAccountUsersFileResourceId,
+            report.invalidAuthCount.toString(),
+            invalidUsersFileResourceId,
             this.api,
             program
         );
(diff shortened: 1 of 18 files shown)


src/data/user-monitoring/two-factor-monitoring/TwoFactorReportD2Repository.ts
```diff
@@ -18,17 +22,40 @@ export class TwoFactorReportD2Repository implements TwoFactorReportRepository {
...
+            "_who"+filenameUserReported,
+            this.api
+        );
+        const invalidUsersFileResourceId = await UserMonitoringFileResourceUtils.saveFileResource(
+            report.invalidAuthList
+                .map(user => {
+                    return user.name + "," + user.id;
+                })
+                .join("\n"),
+            "_invalid"+filenameUserReported,
```
tokland: These two blocks are almost identical, create a private method to abstract it.

src/data/user-monitoring/two-factor-monitoring/TwoFactorReportD2Repository.ts
```diff
@@ -42,7 +69,11 @@ export class TwoFactorReportD2Repository implements TwoFactorReportRepository {
 
     private async push(
         invalidConfigNumber: string,
-        invalidConfigUsers: string,
+        invalidFileReferenceListUsers: string,
+        invalidWhoAccounts: string,
+        invalidFileReferenceWhoAccounts: string,
+        invalidAuth: string,
+        invalidFileReferenceAuth: string,
```
tokland, line 72: All these `invalidXyz` props seem to form a unity. We can create a type and use it as an object, not isolated fields.

src/data/user-monitoring/two-factor-monitoring/TwoFactorReportD2Repository.ts
```diff
@@ -54,10 +85,20 @@ export class TwoFactorReportD2Repository implements TwoFactorReportRepository {
...
+                        if (invalidWhoAccounts == "0") return { dataElement: "", value: "" };
+                        return { dataElement: item.id, value: invalidWhoAccounts };
+                    case dataelement_invalid_who_accounts_list:
+                        if (invalidFileReferenceWhoAccounts == "0") return { dataElement: "", value: "" };
+                        return { dataElement: item.id, value: invalidFileReferenceWhoAccounts };
+                    case dataelement_invalid_auth:
+                        if (invalidAuth == "0") return { dataElement: "", value: "" };
+                        return { dataElement: item.id, value: invalidAuth };
+                    case dataelement_invalid_auth_list:
+                        if (invalidFileReferenceAuth == "0") return { dataElement: "", value: "" };
```
tokland: returning an empty object (`{ dataElement: "", value: "" }`) is a code smell. Instead, let's return undefined, and we remove them with a compact() (with a lodash _(...) chain)

src/data/user-monitoring/two-factor-monitoring/TwoFactorUserD2Repository.ts
```diff
@@ -24,8 +27,38 @@ export class TwoFactorUserD2Repository implements TwoFactorUserRepository {
                 id: user.id,
```
tokland, line 28: Above that, declare the type of the object returned by the map, for better typing:

```ts
return responses["users"].map((user): TwoFactorUser => {
```

src/data/user-monitoring/two-factor-monitoring/TwoFactorUserD2Repository.ts
```diff
@@ -5,15 +5,18 @@ import { TwoFactorUser } from "domain/entities/user-monitoring/two-factor-monito
...
+
+    async getUsersNotInGroupIds(excludeUserGroupIds: string[]): Async<TwoFactorUser[]> {
+        log.info(`Get users not in group: ${excludeUserGroupIds.join(",")}`);
         //todo use d2api filters
+        //We need to check if the program metadata is valid due !in filter is not working propertly in dhis2 2.41
         const responses = await this.api
             .get<Users>(
-                `/users.json?paging=false&fields=*,userCredentials[*]&filter=userGroups.id:in:[${groupIds.join(
+                `/users.json?paging=false&fields=id,username,disabled,externalAuth,userGroups,created,userCredentials[twoFa,twoFactorEnabled]&filter=userGroups.id:!in:[${excludeUserGroupIds.join(
                     ","
```
tokland, line 20: Maybe that's a good time to refactor it to use d2-api

src/domain/entities/user-monitoring/two-factor-monitoring/TwoFactorUserOptions.ts
```diff
@@ -3,4 +3,6 @@ import { NamedRef } from "domain/entities/Base";
 export interface TwoFactorUserOptions {
     pushProgram: NamedRef;
     twoFactorGroup: NamedRef;
+    whoAccountGroup?: NamedRef;
+    exceptionGroup?: NamedRef[]
```
tokland: its NamedRef[], an array, so `exceptionGroups` ?

tokland: Optional props are not ideal on data structure . Better: `prop: Maybe<value>`, so TS reminds us to put some value on initialization.

src/domain/entities/user-monitoring/two-factor-monitoring/TwoFactorUserReport.ts
```diff
@@ -1,6 +1,10 @@
...
 
 export type TwoFactorUserReport = {
-    invalidUsersCount: number;
-    listOfAffectedUsers: NamedRef[];
+    invalidTwoFACount: number;
+    invalidTwoFAList: NamedRef[];
+    invalidWhoCount: number;
+    invalidWhoList: NamedRef[];
+    invalidAuthCount: number;
+    invalidAuthList: NamedRef[];
```
tokland, line 6: This field is the count of the related list, right? It seems redundant, as you'd simply call `.length`. For more complex logic, derived fields are done by creating a class and adding get() properties.

src/domain/usecases/user-monitoring/two-factor-monitoring/RunTwoFactorReportUseCase.ts
```diff
@@ -17,28 +16,107 @@ export class RunTwoFactorReportUseCase {
         private programRepository: UserMonitoringProgramD2Repository
     ) {}
 
-    async execute(): Async<TwoFactorReportResponse> {
+    async execute(shouldDisableInvalidUsers: boolean): Async<TwoFactorReportResponse> {
```
tokland: flags like this should in an options: {...}, otherwise it does not look nice when called -> ie. execute(false)

src/domain/usecases/user-monitoring/two-factor-monitoring/RunTwoFactorReportUseCase.ts
```diff
@@ -17,28 +16,107 @@ export class RunTwoFactorReportUseCase {
...
+            return isEnabled && ((isInWhoGroup && isInAuthGroup) || isNotInWhoOr2FA);
+        });
+
         const report: TwoFactorUserReport = {
-            invalidUsersCount: userItems.length,
-            listOfAffectedUsers: userItems,
+            invalidTwoFACount: invalidTwoFactorUsers.length,
+            invalidTwoFAList: invalidTwoFactorUsers.map(user => {
+                return { id: user.id, name: user.username };
+            }) ?? ["No users found"],
```
tokland: the result of a map, an array, will be an array, never be undefined, so `??` is doing nothing. Also, if it were doing something, TS would complain, as it's not a NamedRef[], the type of the prop.

src/domain/usecases/user-monitoring/two-factor-monitoring/__tests__/TwoFactorTest.data.ts
```diff
@@ -20,15 +8,92 @@ export const NoUsersReport = {
...
+      }],
+};
+
+export const userInTwoFAGroupButWithExternalAuth: TwoFactorUser = {
+    id: "userUid",
+    twoFA: false,
+    disabled: false,
+    username: "username",
+    externalAuth: true,
+    userGroups: [ {
```
tokland: this file (and others) are not prettified -> yarn prettify. (also in editor).

tokland (commented):
In-line comments:

src/data/user-monitoring/two-factor-monitoring/TwoFactorUserD2Repository.ts
idelcano: To maintain compatibility between 2.41 and 2.42, since I need a custom query and I’m getting errors in the user schema, I won’t touch this point for now (unless it actually works and I’m just not aware of it).

I started trying....
        const responses = await this.api.models.users
            .get({
                paging: false,
                fields: {
                    userGroups: { id: true },
                    userCredentials: { twoFa: true, 
                            twoFactorEnabled: true,
                            id: true,
                            username: true,
                            disabled: true,
                            externalAuth: true,
                            created: true, },
                },
                filter: {
                     "userGroups.id": { "!in": excludeUserGroupIds },
                },
            })
            .getData();

But the actual User's scheme don't look compatible between versions
I need do this:
    &fields=id,username,disabled,externalAuth,userGroups,created,twoFa,twoFactorEnabled

The actual scheme looks for the user attributes (like username) inside the usercredentials but userCredentials not exist in 2.42 and we want deploy the script in both versions.

idelcano: All done, except maybe this is a good moment to refactor it to use d2-api.
I left a comment — if you think it’s possible at this stage, we can discuss it in the chat.
Thanks @tokland  for the review!

tokland: Ideally, we should add schemas for v41 and v42 in d2-api, to account for the changes in the user schema, so I think it's ok to leave the manual request for now.

src/domain/entities/user-monitoring/two-factor-monitoring/TwoFactorUserOptions.ts
```diff
@@ -1,8 +1,9 @@
 import { NamedRef } from "domain/entities/Base";
+import { Maybe } from "utils/ts-utils";
 
 export interface TwoFactorUserOptions {
     pushProgram: NamedRef;
     twoFactorGroup: NamedRef;
-    whoAccountGroup?: NamedRef;
-    exceptionGroup?: NamedRef[]
+    whoAccountGroup: Maybe<NamedRef>;
+    exceptionGroup: Maybe<NamedRef[]>;
```
tokland, line 8: [minor] probably `exceptionGroups`, being an array.

tokland (approved):

PR #87, opened 2025-07-16
Feature/improve user monitoring

idelcano:
Required:
https://github.com/EyeSeeTea/d2-api/pull/171

This branch starts in: 
https://github.com/EyeSeeTea/d2-tools/pull/86

This PR includes several important fixes and improvements to the user monitoring logic:

Fixes the issue of sending empty reports: Reports containing only [] (length ≤ 2) are now correctly ignored and not sent.

Changes the way user groups are updated: Now follows the same logic as the user app, avoiding the unintended bug overwrite of sharing settings that was happening with the previous method.

Adds the ability to send explicit empty reports: This helps distinguish between "no invalid users found" and "script did not run" — if no report exists in DHIS2, it means the monitoring did not execute.

Improves some log messages for clarity and consistency.

These updates ensure more accurate reporting, preserve sharing configurations, and improve overall traceability of script execution.

### DIFF
--- a/src/data/user-monitoring/permission-fixer/PermissionFixerUserGroupD2Repository.ts
+++ b/src/data/user-monitoring/permission-fixer/PermissionFixerUserGroupD2Repository.ts
@@ -28,17 +28,27 @@ export class PermissionFixerUserGroupD2Repository implements PermissionFixerUser
     }
     async save(userGroup: PermissionFixerUserGroupExtended, _users: Ref[]): Async<string> {
         try {
-            const response = await this.api.models.userGroups.put(userGroup).getData();
-            if (_(response.errorReports).isEmpty()) {
-                log.info("Users added to minimal group");
-            } else {
-                log.error("Error adding users to minimal group");
-            }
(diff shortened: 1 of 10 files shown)


src/data/user-monitoring/permission-fixer/PermissionFixerUserGroupD2Repository.ts
```diff
@@ -28,17 +29,29 @@ export class PermissionFixerUserGroupD2Repository implements PermissionFixerUser
...
+                op: "add",
+                path: "/users/-",
+                value: { id: userId.id },
+            }));
 
-            log.info(JSON.stringify(response));
+            const response = await this.api
+                .request<string>({
+                    method: "patch" as Method,
+                    url: `/41/userGroups/${userGroup.id}`,
```
cgbautista, line 41: patch method works since v2.37, but even though the version indication is not really used in the backend, having that `41` limits this endpoint to be usable in v2.40 and later.

cgbautista (approved):
Looks good from my side.

PR #88, opened 2025-08-06
Feature/add notifications into user permissions

idelcano:
This PR adds Microsoft Teams notifications to user-permissions, and also fixes existing notifications that were not working due to fetch failing in recent repository/Node.js versions.

What was changed:
Replaced the use of the global fetch API with a direct axios.post(...) call for sending messages to Microsoft Teams.

Why this change was made:
Although fetch is available in modern Node.js versions, it was failing at runtime (fetch failed) in our CLI environment.
To improve stability and proxy support, I switched to axios, which was already used elsewhere in the project and provides better error handling, agent configuration, and proxy control.
Also I use axios due i don't need install other third party packages (it was already used in the repo)

Additionally:

Axios’s internal proxy logic was set to (proxy: false) to avoid conflicts with our custom httpAgent/httpsAgent proxy configuration (it didn't work without it).

### DIFF
--- a/src/data/externalConfig/Namespaces.ts
+++ b/src/data/externalConfig/Namespaces.ts
@@ -1,6 +1,6 @@
 export const d2ToolsNamespace = "d2-tools";
 
-export type Namespace = (typeof Namespace)[keyof typeof Namespace];
+export type Namespace = typeof Namespace[keyof typeof Namespace];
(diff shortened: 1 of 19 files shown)


cgbautista (approved):
Looks good from my side

PR #89, opened 2025-08-26
Feature/fix 242 user monitoring

idelcano:
This script was originally developed for 2.36-2.38, but API changes in 2.41 and now in 2.42 required adjustments.
This PR updates user role handling to use the root user object, which is the recommended approach in 2.42 (and also a good option in 2.41 due is a copy of the userRoles in userCredentials), since user.userCredentials is being removed due to duplicated parameters. 

I’ve kept user.userCredentials.userRoles as a temporary fallback for 2.41 compatibility (especially for pushing users and generating backup files), but it can be removed in the future once we move fully to 2.42.

This branch is already running in our servers (and we could considered funtional testing done).
I think we should do a update of all the instances when all this PR was merged.


cgbautista (approved):
Looks good from my side. We should move away from userCredentials

PR #94, opened 2026-01-28
add sql script generation for deleting obsolete cocs

eperedo:
### :pushpin: References

-   **Issue:** https://app.clickup.com/t/869btxm35

### :memo: Implementation

Generating a sql script for deleting categoryOptionCombos.

```shell
yarn start categoryOptionCombos regenerate \
    --url=https://play.im.dhis2.org/stable-2-40-10 \
    --auth="admin:district" \
     --generate-sql-delete-script
```

Before deleting a `categoryOptionCombo` the script checks for existing data in the `datavalue` and `datavalueaudit` tables. If you have thousands or millions of records, this process can be very slow. 

Adding an index to these tables improves performance significantly:

```sql
CREATE INDEX CONCURRENTLY idx_datavalue_catoptcombo ON datavalue(categoryoptioncomboid);
CREATE INDEX CONCURRENTLY idx_datavalue_attoptcombo ON datavalue(attributeoptioncomboid);
CREATE INDEX CONCURRENTLY idx_datavalueaudit_catoptcombo ON datavalueaudit(categoryoptioncomboid);
CREATE INDEX CONCURRENTLY idx_datavalueaudit_attoptcombo ON datavalueaudit(attributeoptioncomboid);
```

#869btxm35

### DIFF
--- a/README.md
+++ b/README.md
@@ -1117,7 +1117,7 @@ yarn start categoryOptionCombos regenerate \
     --delete-cocs
 ```
 
-Save categoryOptionCombos and generating a sql script for deleting categoryOptionCombos.
+Save categoryOptionCombos and generating a sql script for deleting categoryOptionCombos. if you remove the `persist` flag it will only generate the sql.
 
 ```shell
 yarn start categoryOptionCombos regenerate \
--- a/README.md
+++ b/README.md
@@ -1117,6 +1117,17 @@ yarn start categoryOptionCombos regenerate \
     --delete-cocs
 ```
 
+You can also include a comma separated list if you want to regenerate specific category Combos.
+
+```shell
+yarn start categoryOptionCombos regenerate \
+    --url=https://play.im.dhis2.org/dev \
+    --auth="admin:district" \
+    --persist \
+    --delete-cocs \
+    --cat-combos-ids=id1,id2,id3
+```
+
 Save categoryOptionCombos and generating a sql script for deleting categoryOptionCombos. if you remove the `persist` flag it will only generate the sql.
 
 ```shell
@@ -1124,7 +1135,7 @@ yarn start categoryOptionCombos regenerate \
     --url=https://play.im.dhis2.org/dev \
     --auth="admin:district" \
     --persist \
-    --generate-sql-delete-script
+    --generate-sql-delete-script \
 ```
 
 Before deleting a `categoryOptionCombo` the script checks for existing data in the `datavalue` and `datavalueaudit` tables. If you have thousands or millions of records, this process can be very slow.
--- a/README.md
+++ b/README.md
@@ -1125,7 +1125,7 @@ yarn start categoryOptionCombos regenerate \
     --auth="admin:district" \
     --persist \
     --delete-cocs \
-    --cat-combos-ids=id1,id2,id3
+    --category-combo-ids=id1,id2,id3
 ```
 
 Save categoryOptionCombos and generating a sql script for deleting categoryOptionCombos. if you remove the `persist` flag it will only generate the sql.
@@ -1135,7 +1135,7 @@ yarn start categoryOptionCombos regenerate \
     --url=https://play.im.dhis2.org/dev \
     --auth="admin:district" \
     --persist \
-    --generate-sql-delete-script \
+    --generate-sql-delete-script
 ```
 
 Before deleting a `categoryOptionCombo` the script checks for existing data in the `datavalue` and `datavalueaudit` tables. If you have thousands or millions of records, this process can be very slow.
--- a/README.md
+++ b/README.md
@@ -1105,10 +1105,35 @@ Regenerate categoryOptionCombos from `categoryCombo.categories[].categoryOptions
 
 By default both operations (create+update and delete) are being executed using the `VALIDATE` importMode (dry run). Use the --persist flag to apply changes (create+update) and --delete-cocs to confirm the deletion of obsolete categoryOptionCombos.
 
+In some cases deleting cocs through the API could be really slow. You can pass the `--generate-sql-delete-script` to generate a sql script that you can run directly against the database.
+
+Save and delete categoryOptionCombos
+
 ```shell
 yarn start categoryOptionCombos regenerate \
     --url=https://play.im.dhis2.org/dev \
     --auth="admin:district" \
     --persist \
     --delete-cocs
+```
+
+Save categoryOptionCombos and generating a sql script for deleting categoryOptionCombos.
+
+```shell
+yarn start categoryOptionCombos regenerate \
+    --url=https://play.im.dhis2.org/dev \
+    --auth="admin:district" \
+    --persist \
+    --generate-sql-delete-script
+```
+
+Before deleting a `categoryOptionCombo` the script checks for existing data in the `datavalue` and `datavalueaudit` tables. If you have thousands or millions of records, this process can be very slow.
+
+Adding an index to these tables improves performance significantly:
+
+```sql
+CREATE INDEX CONCURRENTLY idx_datavalue_catoptcombo ON datavalue(categoryoptioncomboid);
+CREATE INDEX CONCURRENTLY idx_datavalue_attoptcombo ON datavalue(attributeoptioncomboid);
+CREATE INDEX CONCURRENTLY idx_datavalueaudit_catoptcombo ON datavalueaudit(categoryoptioncomboid);
+CREATE INDEX CONCURRENTLY idx_datavalueaudit_attoptcombo ON datavalueaudit(attributeoptioncomboid);
 ```
--- a/src/data/CategoryComboD2Repository.ts
+++ b/src/data/CategoryComboD2Repository.ts
@@ -7,28 +7,29 @@ import logger from "utils/log";
 export class CategoryComboD2Repository implements CategoryComboRepository {
     constructor(private api: D2Api) {}
 
-    async getAll(): Promise<CategoryCombo[]> {
-        return this.getAllByPages({ page: 1, pageSize: 100, categoryCombos: [] });
+    async getAll(options: { ids?: Id[] }): Promise<CategoryCombo[]> {
+        return this.getAllByPages({ page: 1, pageSize: 100, categoryCombos: [], ids: options.ids });
     }
 
     private async getAllByPages(options: {
         page: number;
         pageSize: number;
         categoryCombos: CategoryCombo[];
+        ids?: Id[];
     }): Promise<CategoryCombo[]> {
-        const { page, pageSize, categoryCombos } = options;
+        const { page, pageSize, categoryCombos, ids } = options;
 
-        const response = await this.getCategoryCombos({ page, pageSize });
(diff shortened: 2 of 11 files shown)


src/domain/usecases/RegenerateCocsUseCase.ts
```diff
@@ -23,7 +23,9 @@ export class RegenerateCocsUseCase {
         );
 
         await this.saveCocs(categoryComboWithGeneratedCocs, options);
-        await this.deleteCocs(categoryComboWithGeneratedCocs, options);
+        if (options.deleteCocs) {
+            await this.deleteCocs(categoryComboWithGeneratedCocs, options);
```
tokland: method `deleteCocs` also takes option `deleteCocs`, which is confusing, 1) it will only be called when true (as we now have this if). 2) you'd expect to delete, given the method name.

src/scripts/commands/category-option-combos/regenerateCocsCmd.ts
```diff
@@ -52,3 +60,81 @@ function generateJsonReport(categoryCombos: RegenerateCocsUseCaseResult["categor
     writeFileSync(fileName, JSON.stringify(jsonReport, null, 2));
     logger.info(`Report generated: ${fileName}`);
 }
+
+const createDeleteUnusedCategoryOptionCombosSQL = (
```
tokland: [Clean Arch] Having SQL generation logic in the presentation layer (script) is a red flag.

[EDIT] After checking with Jorge, he advises just to put the SQL generation code in the `delete` method of the DHIS2 COC repository. I think any argument you need that does not make sense in the method options, can be specified to the constructor (the sql path, ...)

src/data/CategoryComboD2Repository.ts
```diff
@@ -91,3 +84,49 @@ type D2ApiCategoryCombo = {
...
+
+        const categoryIndexByCocOptionPosition = categoryOptionCombo.categoryOptions.map(categoryOption => {
+            const categoryIndexes = categoryIndexesByOptionId[categoryOption.id] ?? [];
+            const alreadyAssignedCount = assignmentCountByOptionId.get(categoryOption.id) ?? 0;
+            assignmentCountByOptionId.set(categoryOption.id, alreadyAssignedCount + 1);
+
+            if (categoryIndexes.length === 0) return Number.MAX_SAFE_INTEGER;
+
+            return categoryIndexes[Math.min(alreadyAssignedCount, categoryIndexes.length - 1)];
+        });
```
tokland, line 126: [note] This block uses in-place updates while `map` signals "no mutations". The orthodox FP way would be to use `reduce` with a state that holds the cumulative indexes. However, being a local variable, it does not add a lot of value to be too dogmatic, I'd keep it as-is.

src/data/CategoryComboD2Repository.ts
```diff
@@ -91,3 +84,49 @@ type D2ApiCategoryCombo = {
...
+
+        return {
+            ...categoryOptionCombo,
+            categoryOptions: _(categoryOptionCombo.categoryOptions)
+                .map((categoryOption, inputPosition) => ({
+                    categoryOption,
+                    categoryIndex: categoryIndexByCocOptionPosition[inputPosition],
+                    inputPosition,
+                }))
+                .sortBy(item => [item.categoryIndex, item.inputPosition])
```
tokland: This pattern has been on my mind for some time, so I took the time to explore it. According to the docs, it should work: https://lodash.com/docs/4.17.23#sortBy (`_.sortBy(users, ['user', 'age']);` it the equivalent example). However, I was worried if this was sorting things as one would expect (lexycographical order) or there was some string conversion that messes things up. 

```typescript
> var users = [{ 'user': 'fred',   'age': 48 }, { 'user': 'barney', 'age': 31 }, { 'user': 'barney', 'age': 4 }] 
> JSON.stringify(_.sortBy(users, ["user", "age"]))
'[{"user":"barney","age":4},{"user":"barney","age":31},{"user":"fred","age":48}]' 
> JSON.stringify(_.sortBy(users, [u => u.user, u => u.age]))
'[{"user":"barney","age":4},{"user":"barney","age":31},{"user":"fred","age":48}]' 
> JSON.stringify(_.sortBy(users, u => [u.user, u.age]))
'[{"user":"barney","age":31},{"user":"barney","age":4},{"user":"fred","age":48}]' 
```

So the code that works is passing mappers as an array (with string -we do not use that, not type-safe, or an arrow function), but not returning an array, as JS compares them as strings.

Also, in addition of fixing, the code if you can tweak the spec so we have a RED before going to GREEN, great!

src/data/CategoryOptionComboDeleteD2SqlExporter.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    exportDeleteScript(cocIds: string[]): void {
+        const sqlQuery = this.createDeleteUnusedCategoryOptionCombosSQL(cocIds);
+        this.writeToDisk(sqlQuery);
+    }
+
+    private createDeleteUnusedCategoryOptionCombosSQL(ids: Id[], batchSize: number = 500): string {
+        const batches = Array.from({ length: Math.ceil(ids.length / batchSize) }, (_, i) =>
+            ids.slice(i * batchSize, (i + 1) * batchSize)
+        );
```
tokland: Is the same than _.chunk ?

src/data/CategoryOptionComboDeleteD2SqlExporter.ts
```diff
@@ -0,0 +1,84 @@
...
+DELETE FROM categorycombos_optioncombos 
+WHERE categoryoptioncomboid IN (SELECT categoryoptioncomboid FROM temp_ids_batch);
+
+DELETE FROM categoryoptioncombo 
+WHERE categoryoptioncomboid IN (SELECT categoryoptioncomboid FROM temp_ids_batch);
+
+DROP TABLE temp_uids_batch;
+DROP TABLE temp_ids_batch;
+
+`;
```
tokland: [subjective] To avoid a long multi-line variable in the middle of a function/method, we can move it to a constant/instance value and use lodash template to interpolate when used, what do you think?

src/data/CategoryOptionComboDeleteD2SqlExporter.ts
```diff
@@ -0,0 +1,84 @@
...
+        });
+
+        return `BEGIN;${batchStatements.join("\n")}COMMIT;`.trim();
```
tokland: Very minor, but id' write an a multiline array with a final join., something like this

return [
            "BEGIN;", //
            ...batchStatements,
            "COMMIT;",
        ].join("\n");

(in the hope that it's more readable)

src/domain/usecases/RegenerateCocsUseCase.ts
```diff
@@ -110,7 +122,7 @@ export class RegenerateCocsUseCase {
 
                 return {
                     regeneratedCoc: RegeneratedCoc.create({
-                        id: getUid(combinationKey, categoryCombo.id),
+                        id: categoryComboId,
```
tokland: So `categoryOptionComboId`?

src/domain/repositories/CategoryOptionComboDeleteExporter.ts
```diff
@@ -0,0 +1,5 @@
+import { Id } from "domain/entities/Base";
+
+export interface CategoryOptionComboDeleteExporter {
+    exportDeleteScript(cocIds: Id[]): void;
```
tokland: Proposal: instead of a side-effect (saving the file), return here a string. And make the use case return also that string somewhere in the response object. And it's the script that writes the final file from that string. What do you think?

tokland (commented):
In-line comments:

src/data/CategoryComboD2Repository.ts
```diff
@@ -7,28 +7,29 @@ import logger from "utils/log";
 export class CategoryComboD2Repository implements CategoryComboRepository {
     constructor(private api: D2Api) {}
 
-    async getAll(): Promise<CategoryCombo[]> {
-        return this.getAllByPages({ page: 1, pageSize: 100, categoryCombos: [] });
+    async getAll(options: { ids?: Id[] }): Promise<CategoryCombo[]> {
+        return this.getAllByPages({ page: 1, pageSize: 100, categoryCombos: [], ids: options.ids });
```
tokland: [Minor, subjective] Having getAll with an optional ids filter feels a bit confusing. Maybe it would be clearer to have two separate methods: getAll() and getByIds(ids)?

src/scripts/commands/category-option-combos/regenerateCocsCmd.ts
```diff
@@ -72,3 +83,9 @@ function generateJsonReport(categoryCombos: RegenerateCocsUseCaseResult["categor
     writeFileSync(fileName, JSON.stringify(jsonReport, null, 2));
     logger.info(`Report generated: ${fileName}`);
 }
+
+function writeSqlScriptToDisk(sqlScript: string, fileName: string): void {
+    writeFileSync(fileName, sqlScript);
+    logger.info(`SQL generated: ${fileName}`);
+    logger.info(`You can execute the sql with d2-docker: d2-docker run-sql ${fileName}`);
```
tokland: we may have N instances running, I'd add the explicit image option

README.md
```diff
@@ -1117,14 +1117,25 @@ yarn start categoryOptionCombos regenerate \
...
 
+You can also include a comma separated list if you want to regenerate specific category Combos.
+
+```shell
+yarn start categoryOptionCombos regenerate \
+    --url=https://play.im.dhis2.org/dev \
+    --auth="admin:district" \
+    --persist \
+    --delete-cocs \
+    --cat-combos-ids=id1,id2,id3
```
tokland: Proposal: --category-combo-ids (no abbreviation, plural only the last term)

README.md
```diff
@@ -1117,14 +1117,25 @@ yarn start categoryOptionCombos regenerate \
...
+
 Save categoryOptionCombos and generating a sql script for deleting categoryOptionCombos. if you remove the `persist` flag it will only generate the sql.
 
 ```shell
 yarn start categoryOptionCombos regenerate \
     --url=https://play.im.dhis2.org/dev \
     --auth="admin:district" \
     --persist \
-    --generate-sql-delete-script
+    --generate-sql-delete-script \
```
tokland: do we need it?

src/data/CategoryComboD2Repository.ts
```diff
@@ -39,6 +40,8 @@ export class CategoryComboD2Repository implements CategoryComboRepository {
                 },
                 page: options.page,
                 pageSize: options.pageSize,
+                // TODO: change to request in chunks when ids are provided
+                filter: options.ids ? { id: { in: options.ids } } : undefined,
```
tokland: Is this TODO to do now, as we have the ids?

tokland (commented):
Some very minor things:

MiquelAdell (approved):

PR #95, opened 2026-01-30
[feature]  Delete data values that belong to given data elements

nshandra:
### :pushpin: References

Issue: [create (or reuse) a script to delete all data for a list of data elements for all periods and org units](https://app.clickup.com/t/869bxqbjw)

### :memo: Implementation

Bulk delete data values that belong to the DEs in the provided CSV. This file is expected to be single column with header id.

It splits the provided DEs in chunks of 200 (to not hit the url limit), retrieves `limit`amount of dataValues and posts them with DELETE strategy. Repeats untill no more data is retrieved and starts with the next DEs block.

The deleted values can be optionally stored in backup files.

#869bxqbjw


MiquelAdell:
- after runing, on the branch
```
nvm use
yarn
yarn build
```

I'm getting:
``` 
 yarn start datavalues bulk-delete --url='https://dev.user:password@dev.eyeseetea.com/who-dev-41' dataElements.csv 
yarn run v1.22.22
$ node dist/index.js datavalues bulk-delete '--url=https://dev.user:password@dev.eyeseetea.com/who-dev-41' dataElements.csv
node:internal/errors:496
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_INVALID_URL]: Invalid URL
    at new NodeError (node:internal/errors:405:5)
    at new URL (node:internal/url:676:13)
    at p (/Users/miqueladell/code/d2-tools/dist/index.js:1:266140)
    at t.getD2Api (/Users/miqueladell/code/d2-tools/dist/index.js:1:266528)
    at Object.handler (/Users/miqueladell/code/d2-tools/dist/index.js:1:230212)
    at Object.run (/Users/miqueladell/code/d2-tools/node_modules/cmd-ts/dist/cjs/command.js:150:41)
    at async Object.run (/Users/miqueladell/code/d2-tools/node_modules/cmd-ts/dist/cjs/subcommands.js:149:32)
    at async Object.run (/Users/miqueladell/code/d2-tools/node_modules/cmd-ts/dist/cjs/subcommands.js:149:32)
    at async runSafely (/Users/miqueladell/code/d2-tools/node_modules/cmd-ts/dist/cjs/runner.js:38:24)
    at async run (/Users/miqueladell/code/d2-tools/node_modules/cmd-ts/dist/cjs/runner.js:10:20) {
  input: 'https://dev.user:password@dev.eyeseetea.com/who-dev-41',
  code: 'ERR_INVALID_URL'
}

Node.js v18.20.8
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Is there something I'm missing?


- can we have a dry run option? and make that default?
Check how other d2-tools do that if possible for consistency

- I was getting an error on some passwords with some characters. I fixed it by singlequoting the url instead of doble quoting it
can you please update that on the readme? 

thanks!

nshandra:
@MiquelAdell Regarding the `url` param, i switched to a `url` and `auth` method that works better with real, complex passwords.
I suspect the error stack you posted is also caused by the old auth method.

Sorry for that.

MiquelAdell (approved):

src/domain/repositories/BulkDeleteDEsRepository.ts
```diff
@@ -0,0 +1,3 @@
+export interface BulkDeleteDEsRepository {
+    getDEsToDelete(csv: string): Promise<string[]>;
+}
```
tokland: EOL missing, maybe some editor setting to enable? (also in other files)

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
```diff
@@ -0,0 +1,126 @@
...
+    dryRun: boolean;
+}
+
+export class BulkDeleteDataValuesUseCase {
+    private dryRun = false;
+    private backupFolder: string | undefined;
+    private deChunkSize = 200;
+
+    constructor(
+        private bulkDeleteRepository: BulkDeleteDEsRepository,
```
tokland: We used this pattern for some time, but Jorge clarified that this kind of operations do not belong to a repository. The idea:

- It's the script (as the presentation layer) that has to convert that input to some value  ready to be passed to the use case. In this case, the script would read the CSV and get the list of IDs and pass within BulkDeleteDataValuesOptions. You can keep the current interface and implementation, and use it from the presentation, just rename it, as it's not a repository.

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
```diff
@@ -0,0 +1,126 @@
...
+        private orgUnitRepository: OrgUnitRepository,
+        private dataValuesRepository: DataValuesRepository
+    ) {}
+
+    async execute(options: BulkDeleteDataValuesOptions) {
+        const { backupFolder, dryRun } = options;
+        if (backupFolder) this.validatePath(backupFolder);
+
+        this.dryRun = dryRun;
+        this.backupFolder = backupFolder;
```
tokland: This pattern breaks FP immutability. Properties of the instance should only be set in the constructor. Anywhere else, they should be kept immutable. We have two typical solutions here:

1) pass the options argument to the methods that need of it.
2) create a separate internal class where these props are passed in the constructor, then they are ready to be used in any method of the class.

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
```diff
@@ -0,0 +1,126 @@
...
+
+        const dataElementIds = await this.bulkDeleteRepository.getDEsToDelete(options.dataElementsFile)
+            .then(ids => {
```
tokland: [best practices] as a general rule, we should avoid mixing await and then style. We can await here.

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
```diff
@@ -0,0 +1,126 @@
...
+                if (ids.length === 0) {
+                    log.error("CSV empty or missing headers.");
+                    process.exit(0);
```
tokland: A use case should not process.exit, that's a script-only operation. Use cases are agnostic. The use case should return (in the return value) all the script needs to know about how the operation went. And then it's the script that executes process.exit, with that info.

tokland: 0 as return value means success, but that's an error branch, right?

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
```diff
@@ -0,0 +1,126 @@
...
+                dataValues: dataValues,
+                dryRun: this.dryRun,
+            });
+        } catch (error) {
+            log.error(`Error deleting data values in batch ${backupIndex}: ${(error as Error).message}`);
+            process.exit(1);
+        }
+
+        if (this.dryRun) {
+            log.info("Dry run mode - stopping after first batch.");
```
tokland: That's weird for a dry-run, which typically does the full operation (all the batches), just that the values are not actually deleted.

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
```diff
@@ -0,0 +1,126 @@
...
+                return _.uniq(ids);
+            });
+        log.info(`Data elements to delete count: ${dataElementIds.length}`);
+
+        const rootOrgUnit = await this.orgUnitRepository.getRoot();
+        const deChunks = _.chunk(dataElementIds, this.deChunkSize);
+
+        let backupIndex = 1;
+        const lastIndex = deChunks.length - 1;
+        for (const [index, dataElements] of deChunks.entries()) {
```
tokland: [namig] To be more exact: `dataElementIds`.

src/scripts/commands/dataValues.ts
```diff
@@ -254,3 +257,40 @@ const monitoringDataValues = command({
...
+const bulkDeleteDataValuesCmd = command({
+    name: "bulk-delete",
+    description: "Bulk delete data values based on data element CSV",
+    args: {
+        ...getApiUrlOptions(),
+        limit: option({
+            type: number,
+            long: "limit",
+            description: "Number of data values to delete in each batch (default: 100000)",
+            defaultValue: () => 30000,
```
tokland: default values do not match. We can use a variable so they never are out of sync.

src/scripts/commands/dataValues.ts
```diff
@@ -254,3 +257,40 @@ const monitoringDataValues = command({
...
+        }),
+        backupFolder: option({
+            type: optional(string),
+            long: "backup-folder",
+            description: "Folder for backups, leave empty to disable. Will be stored as bulk-delete-backup-<batch>-<timestamp>.json",
+        }),
+        dryRun: flag({
+            long: "dry-run",
+            description: "Perform delete in dry run mode. Limit MUST be higher than the number of data values to delete per DE batch.",
```
tokland: Second part: "Limit MUST..." is not for this option, right?

README.md
```diff
@@ -370,6 +370,42 @@ $ yarn start datavalues monitoring-values \
...
+
+Finally if the target instance has limited resources or high usage consider lowering the limit.
+
+Script launch options:
+
+```shell
+OPTIONS:
+  --url <str>           - http[s]://[USERNAME:PASSWORD@]HOST:PORT
+  --auth <value>        - USERNAME:PASSWORD [optional]
+  --limit <number>      - Number of data values to delete in each batch (default: 30000) [optional]
```
tokland: The "data values per batch" setting is unclear to users without checking the code (it chunks dataElements -size 200- and applies a limit). Maybe we can rephase to make clear the limit is by request or something like this.

A the end, the real problem here is: users cannot know if a deletion will be total or partial.

While we would prefer (I guess) to delete all values, the lack of pagination on the dataValueSets endpoint makes this difficult. A generic solution would require complex dynamic request partitioning (by orgUnit, periods, etc.), which is not trivial.

Conclusion: I guess the current script works for our specific needs. We can keep it as is and address the partial-deletion limitation only if it becomes a priority.

src/data/BulkDeleteDEsCsvRepository.ts
```diff
@@ -0,0 +1,29 @@
...
+    async getDEsToDelete(csv: string): Promise<string[]> {
+        return new Promise((resolve, reject) => {
+            const dataElementIds: string[] = [];
+
+            fs.createReadStream(csv, "utf8")
+                .pipe(new CsvReadableStream({ asObject: true, trim: true }))
+                .on("data", rawRow => {
+                    const row = rawRow as unknown as Ref;
+                    if (!row.id) return;
+                    if (row) {
```
tokland: if (row.id) { ... } should be enough

tokland (commented):
Feature worked fine in my tests.

About the code: Some in-line comments. I am not sure how much time we have left in this task for refactors, probably the comments about the limit/pagination are to simply be skipped. The rest should be doable.

src/scripts/commands/dataValues.ts
```diff
@@ -254,3 +257,40 @@ const monitoringDataValues = command({
...
+            description: "CSV file with data element IDs and id header",
+        }),
+    },
+    handler: async args => {
+        const api = getD2ApiFromArgs(args);
+        const bulkDeleteRepository = new BulkDeleteDEsCsvRepository();
+        const orgUnitRepository = new OrgUnitD2Repository(api);
+        const dataValuesRepository = new DataValuesD2Repository(api);
+
+        new BulkDeleteDataValuesUseCase(bulkDeleteRepository, orgUnitRepository, dataValuesRepository).execute(args);
```
tokland: await missing here?

README.md
nshandra: I expanded the README with a explanation of how the script works, as its not clear without checking the code. The deletion will be total (assuming no errors). 
Limit is batch-size now. The dataElements chunks are DE groups.

Extract from README about this:

> #### Notes about DE groups and batch size
> 
> -   DE groups: Given that the user can provide a high number of DEs via the CSV file, the script process them in groups of up to 200 DEs at a time.
> 
> -   Batches: Within each group, data values are fetched and deleted in batches of up to `batch-size` values per API call. This repeats until there are no more values to delete for that group.
> 
> Example log to illustrate how it works:
> 
> ```log
> [INFO TIMESTAMP] Data elements to delete count: 300
> [INFO TIMESTAMP] Processing data element group 1 of 2 (DEs: 200)
> [INFO TIMESTAMP] Data values fetched for deletion in batch 1: 30000
> [INFO TIMESTAMP] POST /dataValues response: {"imported":0,"updated":0,"ignored":0,"deleted":30000}
> [INFO TIMESTAMP] Data values fetched for deletion in batch 2: 1000
> [INFO TIMESTAMP] POST /dataValues response: {"imported":0,"updated":0,"ignored":0,"deleted":1000}
> [INFO TIMESTAMP] Processing data element group 2 of 2 (DEs: 100)
> [INFO TIMESTAMP] Data values fetched for deletion in batch 3: 20000
> [INFO TIMESTAMP] POST /dataValues response: {"imported":0,"updated":0,"ignored":0,"deleted":20000}
> [INFO TIMESTAMP] All data values have been deleted.
> Done in RUNTIME.
> ```
> 
> When using the `dry-run` option, the script will stop after the first batch. This is because the dataValueSets endpoint has no pagination and it always return the first N data values (N being `batch-size` here), so there is no easy way of getting the next batch.

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
nshandra: Yes, it is a bit janky. Because `dataValueSets` lack pagination we cant perform all the batches without deleting.
Without dry-run the script runs the `deleteDataValuesRecursively` function until it cant get more values for the current DEs batch (fetched count < limit). This doesn't work without deleting hence the stop after the first iteration.

Related: https://github.com/EyeSeeTea/d2-tools/pull/95#discussion_r3100332107

src/scripts/commands/dataValues.ts
nshandra: It is, when the script deletes the first N dataValues it can get the next N values from dataValueSet. With dry-run it cant work recursively as it would be stuck in the first batch, so to test all the values for the group the batch-size (limit) has to be higher.  

As mentioned in: https://github.com/EyeSeeTea/d2-tools/pull/95#discussion_r3100338918 its a bit janky, but i didn't find a way of making a working dry-run without `dataValueSet` pagination.

src/domain/repositories/BulkDeleteDEsRepository.ts
nshandra: Looks like sometimes VSCode doesn't apply formatting on save. I will keep an eye on this.

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
nshandra: Moved the CSV read and empty check to the script section. Removed the repository.

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
nshandra: Switched to option 1.

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
nshandra: Noted.

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
nshandra: Moved the process.exit into an exception handler in the script.

nshandra: True, is an exception now.

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
nshandra: As `dataElementIds` is used earlier, switched to `deIds`.

src/scripts/commands/dataValues.ts
nshandra: Switched to a variable for the default.

src/data/BulkDeleteDEsCsvRepository.ts
nshandra: Changed.

src/scripts/commands/dataValues.ts
nshandra: Yes, it was missing.

README.md
```diff
@@ -381,28 +381,60 @@ The values to be deleted can be stored in JSON files in case a backup is needed.
 Given the size of the potential API error response consider storing the script output in a file:
 
 ```shell
-yarn start datavalues bulk-delete --url='http://localhost:8080' --auth='admin:district' --limit=1000  deToDelete.csv &> run.log
+yarn start datavalues bulk-delete --url='http://localhost:8080' --auth='admin:district' --batch-size=1000  deToDelete.csv &> run.log
+# or
+yarn start datavalues bulk-delete --url='http://localhost:8080' --auth='admin:district' --batch-size=1000  deToDelete.csv 2>&1 | tee run.log
```
tokland: [minor] two spaces: `1000  deToDelete.csv `. 

For the tee-all, we can write `|& tee run.log` (not sure if we need to show how to redirect output to file in this specific command, but it won't hurt).

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
```diff
@@ -1,118 +1,122 @@
...
+import { Logger } from "domain/logger/Logger";
 
-import { BulkDeleteDEsRepository } from "domain/repositories/BulkDeleteDEsRepository";
 import { DataValuesRepository } from "domain/repositories/DataValuesRepository";
 import { OrgUnitRepository } from "domain/repositories/OrgUnitRepository";
 import { DataValue } from "domain/entities/DataValue";
 
-
 interface BulkDeleteDataValuesOptions {
     dataElementsFile: string;
```
tokland: Is this field still used?

src/data/DataValuesD2Repository.ts
```diff
@@ -42,10 +42,16 @@ export class DataValuesD2Repository implements DataValuesRepository {
         return res.dataValues;
     }
 
-    private async postDataValueSet(options: { dataValues: DataValueToPost[], postParams?: DataValueSetsPostParams, showCount?: boolean }): Async<void> {
+    private async postDataValueSet(options: {
+        dataValues: DataValueToPost[];
+        postParams?: DataValueSetsPostParams;
+        showCount?: boolean;
```
tokland, line 48: [note] No need to change, very minor, but options like showCount somewhat pollute a public method of a repository. Such options fit better in the constructor.

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
```diff
@@ -1,118 +1,122 @@
...
                 dataValues: dataValues,
-                dryRun: this.dryRun,
+                dryRun: dryRun,
             });
         } catch (error) {
-            log.error(`Error deleting data values in batch ${backupIndex}: ${(error as Error).message}`);
-            process.exit(1);
+            throw new Error(
+                `Error deleting data values in batch ${backupIndex}: ${(error as Error).message}`
+            );
```
tokland, line 94: [note] When re-throwing exceptions, there is a nice pattern, passing ` { cause: error }` as the second argument, we then don't lose the original error stacktrace.

src/scripts/commands/dataValues.ts
```diff
@@ -286,11 +298,49 @@ const bulkDeleteDataValuesCmd = command({
...
+        } catch (error) {
+            console.error((error as Error).message);
+            process.exit(1);
+        }
     },
 });
 
+async function readDataElementsFile(csv: string): Promise<string[]> {
+    if (!fs.existsSync(csv) || !fs.statSync(csv).isFile()) {
+        throw new Error(`Cant find file: ${csv}`);
```
tokland: Cant -> Can't. Or: File not found. 

And maybe: csv -> csvPath to make very clear it's not contents but a path.

tokland (commented):
Some minor in-line comments:

README.md
nshandra: The redirect helps making making the log more readable, especially in case of issues. Fixed the double space.

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
nshandra: No, deleted.

src/scripts/commands/dataValues.ts
nshandra: Changed.

src/domain/usecases/BulkDeleteDataValuesUseCase.ts
nshandra: Oh, nice. For this script it only re-prints the info of [d2-api:request] exception, so no point in duplicating a large message. But good to know anyway.

tokland (approved):
Thx, all good as far as I can see!

PR #96, opened 2026-03-18
Fix/disable users server freeze

idelcano:
https://app.clickup.com/t/869c8hupk

This PR implements request batching for the user disabling process within RunTwoFactorReportUseCase.

The Problem: Previously, attempting to disable a large number of users simultaneously caused the server to crash or time out due to an excessive number of concurrent requests.

The Solution: I introduced a batching strategy with a limit of 50 users per block. This ensures system stability while keeping the implementation as surgical as possible, minimizing changes to the existing logic and response contract.

### DIFF
--- a/src/domain/usecases/user-monitoring/two-factor-monitoring/RunTwoFactorReportUseCase.ts
+++ b/src/domain/usecases/user-monitoring/two-factor-monitoring/RunTwoFactorReportUseCase.ts
@@ -124,9 +124,7 @@ export class RunTwoFactorReportUseCase {
                 const batchResponse = await this.userRepository.disableUsers(userIdsBatch);
                 const batchResults = JSON.parse(batchResponse);
 
-                if (Array.isArray(batchResults)) {
-                    disableResults.push(...batchResults);
-                }
+                disableResults.push(...batchResults);
             } catch {
                 disableResults.push({
                     userId: "unknown",
--- a/src/domain/usecases/user-monitoring/two-factor-monitoring/RunTwoFactorReportUseCase.ts
+++ b/src/domain/usecases/user-monitoring/two-factor-monitoring/RunTwoFactorReportUseCase.ts
@@ -91,12 +91,13 @@ export class RunTwoFactorReportUseCase {
         const saveResponse = await this.reportRepository.save(programMetadata, report);
         if (shouldDisableInvalidUsers) {
             if (invalidTwoFactorUsers.length > 0) {
-                const disableResponse = await this.userRepository.disableUsers(
-                    invalidTwoFactorUsers.map(user => user.id)
+                const disableResponse = await this.disableUsersInBatches(
+                    invalidTwoFactorUsers.map(user => user.id),
+                    50
                 );
                 return {
                     message: saveResponse,
-                    disableUsersMessage: JSON.stringify(disableResponse),
+                    disableUsersMessage: disableResponse,
                     report,
                 };
             } else {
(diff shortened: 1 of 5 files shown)


src/domain/usecases/user-monitoring/two-factor-monitoring/RunTwoFactorReportUseCase.ts
```diff
@@ -91,12 +91,13 @@ export class RunTwoFactorReportUseCase {
         const saveResponse = await this.reportRepository.save(programMetadata, report);
         if (shouldDisableInvalidUsers) {
             if (invalidTwoFactorUsers.length > 0) {
-                const disableResponse = await this.userRepository.disableUsers(
-                    invalidTwoFactorUsers.map(user => user.id)
+                const disableResponse = await this.disableUsersInBatches(
+                    invalidTwoFactorUsers.map(user => user.id),
+                    50
```
MatiasArriola: Let's avoid using magic numbers, either extract to a constant or use a parameter with a sensible default value (that could also be extracted to a constant)

MatiasArriola (changes requested):
This PR should work fine. However here are some notes, mostly related to clean architecture:

1. (preexisting) `disableUsers` is implemented in `TwoFactorUserD2Repository.ts`, but should be declared in the abstract interface `TwoFactorUserRepository.ts`. This would add flexibility if tomorrow we want to switch to other Repository implementation
2. (preexisting) `disableUsers` returns a string, which in fact is a `JSON.stringify`. Repositories usually return Domain objects or in some cases void or some object signalling the operation outcome. We should avoid doing `JSON.stringify` in the repo and then `JSON.parse` in the useCase, when we could return the object directly and type it accordingly.
3. Since we are doing the batching/chunking of requests because of a transport-level limitation, I think we could move the chunking logic into inside the `disableUsers` repository method. So we keep business-related logic in the useCase and don't mix it with specific concerns like how DHIS handles disabling users. If tomorrow that limitation goes away, you would only modify the repository implementation.

I think at least we should fix the magic number, and if possible address these other points. 
Let me know if you have any concern or need help and we can discuss it.

idelcano:
> This PR should work fine. However here are some notes, mostly related to clean architecture:
> 
> 1. (preexisting) `disableUsers` is implemented in `TwoFactorUserD2Repository.ts`, but should be declared in the abstract interface `TwoFactorUserRepository.ts`. This would add flexibility if tomorrow we want to switch to other Repository implementation
> 2. (preexisting) `disableUsers` returns a string, which in fact is a `JSON.stringify`. Repositories usually return Domain objects or in some cases void or some object signalling the operation outcome. We should avoid doing `JSON.stringify` in the repo and then `JSON.parse` in the useCase, when we could return the object directly and type it accordingly.
> 3. Since we are doing the batching/chunking of requests because of a transport-level limitation, I think we could move the chunking logic into inside the `disableUsers` repository method. So we keep business-related logic in the useCase and don't mix it with specific concerns like how DHIS handles disabling users. If tomorrow that limitation goes away, you would only modify the repository implementation.
> 
> I think at least we should fix the magic number, and if possible address these other points. Let me know if you have any concern or need help and we can discuss it.

Thanks for the feedback @MatiasArriola ! 
I went through it and realized it wouldn't take much, so I went ahead and implemented everything, including the output formatting due was weird (the script already send the report to a event in dhis2 too with the details, but better have a good summary of the script).

MatiasArriola (approved):
Thanks @idelcano !!

These changes improved the code a lot. Only thing, now that the use case correctly uses repository interfaces, there are some unused imports in `RunTwoFactorReportUseCase.ts` that are just leftovers.
I'm approving anyway, just make sure to clean those up.

PR #97, opened 2026-03-20
Feature/disable twfa users since creationdate

idelcano:
📌 References
Issue: Closes https://app.clickup.com/t/869c8hupk
📝 Implementation
Problem:
We have been disabling users weekly because they weren't activating their 2FA. However, with the new release, we can now force 2FA activation. This means we no longer need to disable accounts so quickly; instead, we can target only those users who haven't set up 2FA within X months of their account creation.

Solution:
Added disableAfterMonths dataset setting.
Added the --filtered-by-month CLI flag to run-2fa-reporter. Although it doesn't accept a specific month value via CLI, it acts as a toggle to enable the filtering logic. This allows us to maintain the specific threshold in the Datastore for easy updates while giving us the flexibility to enable or disable filtering at the server level.

When enabled, only invalid users whose account was created >= [disableAfterMonths] ago are reported/disabled.
Throws an error if --filtered-by-month is used but is not configured in the datastore (invalid config).

### DIFF
--- a/src/domain/entities/user-monitoring/two-factor-monitoring/TwoFactorUser.ts
+++ b/src/domain/entities/user-monitoring/two-factor-monitoring/TwoFactorUser.ts
@@ -1,5 +1,6 @@
 import { Id, NamedRef } from "domain/entities/Base";
 import { Timestamp } from "domain/entities/Date";
+import { getMonthsDiff } from "domain/entities/DateTime";
 
 export interface TwoFactorUser {
     id: Id;
@@ -21,12 +22,9 @@ export function filterByCreationDate(
         );
     }
 
-    const now = new Date();
+    const now = new Date().toISOString();
     return users.filter(user => {
-        const createdDate = new Date(user.created);
-        const monthsDiff =
-            (now.getFullYear() - createdDate.getFullYear()) * 12 +
-            (now.getMonth() - createdDate.getMonth());
+        const monthsDiff = getMonthsDiff(user.created, now);
         return monthsDiff >= disableAfterMonths;
     });
 }
--- a/src/domain/entities/user-monitoring/two-factor-monitoring/TwoFactorUser.ts
+++ b/src/domain/entities/user-monitoring/two-factor-monitoring/TwoFactorUser.ts
@@ -1,5 +1,5 @@
 import { Id, NamedRef } from "domain/entities/Base";
-import { Maybe } from "utils/ts-utils";
+import { Timestamp } from "domain/entities/Date";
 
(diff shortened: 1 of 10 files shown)


idelcano:
@MatiasArriola this PR starts from: 
https://github.com/EyeSeeTea/d2-tools/pull/96

src/domain/entities/user-monitoring/two-factor-monitoring/TwoFactorUser.ts
```diff
@@ -7,4 +8,26 @@ export interface TwoFactorUser {
     disabled: boolean;
     externalAuth: boolean;
     userGroups: NamedRef[];
+    created: Maybe<string>;
```
MatiasArriola: Is it possible that a `TwoFactorUser` exists without the `created` prop? if not, I would try making it required, leaving it `created: string;`. 
Also could use existing `Timestamp` type, which is just an alias to `string`

src/domain/entities/user-monitoring/two-factor-monitoring/TwoFactorUser.ts
```diff
@@ -7,4 +8,26 @@ export interface TwoFactorUser {
...
+        );
+    }
+
+    const now = new Date();
+    return users.filter(user => {
+        if (!user.created) return false;
+        const createdDate = new Date(user.created);
+        const monthsDiff =
+            (now.getFullYear() - createdDate.getFullYear()) * 12 +
+            (now.getMonth() - createdDate.getMonth());
```
MatiasArriola: Comparing `2026-04-01` with a user created `2026-03-31` would give a diff of 1 month. This could be expected behavior, but could be better documented with some unit tests having edge cases like that.

Maybe we can extract this dates comparison logic outside for better reuse. I see there is already a `DateTime.ts` file in the project. 

Also luxon is listed in the dependencies and has [utilities](https://moment.github.io/luxon/#/math?id=diffs) for comparison.

MatiasArriola (approved):
Thanks @idelcano! Looks good, I'm approving and leaving some non-blocking comments.

src/domain/entities/user-monitoring/two-factor-monitoring/TwoFactorUser.ts
idelcano: Hi @MatiasArriola , thanks for the review! 
I applied both changes—they were quick to do and definitely make everything look better. Thanks!

MatiasArriola (approved):
Good updates, Thanks @idelcano!

PR #99, opened 2026-06-04
Add dependency track githubaction

idelcano:
📌 References
This PR enables repository access to the master branch and the secret, allowing the Dependency-Track dependency check to run.

Issue: Closes #?
https://app.clickup.com/t/4528615/869b6mend

📝 Implementation
Added the DTRACK_API_KEY secret to the shared workflow call so that the Dependency-Track jobs are no longer skipped. From now on, every push to master will generate a BOM and upload it to Dependency Track for vulnerability analysis.

📹 Screenshots/Screen capture
N/A

🔥 Testing
After merging, check the Actions tab to confirm the Dependency-Track jobs run successfully. Then open the Dependency-Track project link from the workflow summary to review the dependency analysis. If vulnerabilities are found, triage them there and resolve or suppress as appropriate.


anagperal (changes requested):
thanks @idelcano !

Two changes to the CI config here, so this PR finishes adopting the shared workflow rather than half-adopting it.

**`main.yml`: `pull_request` is no filtered by branch.** That filter matches the _base_ branch, so any PR targeting a feature branch got no automated feedback at all — no tests, no CodeQL, no dependency scan — and every analysis had to be dispatched by hand. `push` stays limited to `master`/`development`.

```
 on:
      push:
        branches: ["master", "development"]
      pull_request:
      workflow_dispatch:
```

**Removed `codeql.yml`.** The shared workflow already runs CodeQL as a job, so it was analysing the same commit twice.

`test.yml` stays — it runs on every push with no PR open, which the shared workflow does not cover.

idelcano:
Hi @anagperal 
I've applied the changes. It seems CodeQL is equivalent, and I've removed master/development, although you will confirm this in a technical meeting due all the branch was merged using only master/dev as github actions triggers.

anagperal:
> Hi @anagperal I've applied the changes. It seems CodeQL is equivalent, and I've removed master/development, although you will confirm this in a technical meeting due all the branch was merged using only master/dev as github actions triggers.

Thanks @idelcano ! Yes, we'll discuss in the next technical decisions meeting if we want to maintain run only PRs with base branch master/main/development or change it to run on every PR regardless of base branch, in general

PR #104, opened 2026-06-26
feat: add --created-date-overrides-file flag to 2FA reporter

idelcano:
issue: https://app.clickup.com/t/4528615/869dw5eb9
  Extends the `run-2fa-reporter` command with a new `--created-date-overrides-file` flag.

  When `--filtered-by-month` is active, the reporter only disables invalid 2FA users whose
  account is older than `disableAfterMonths` (from the datastore). This new flag lets you
  override the creation date used for that check on a per-user basis: instead of the real
  DHIS2 creation date, the script uses the date provided in a JSON file.

  This is useful when an account exists in DHIS2 from long ago but should be treated as
  recently created for this check — e.g. to protect a re-onboarded user from being disabled.

  The override is a pure date substitution and works in both directions: a user given a
  recent date is protected, and a user given an old-enough date is still disabled.
  
  Also adds the previously-missing README documentation for the existing `--filtered-by-month` flag.

  ## Implementation notes

  - `UserDateOverride` entity and the pure `applyUserDateOverrides` function live in the
    domain layer next to `filterByCreationDate`.
  - The override file is parsed and validated at the CLI boundary (`parseUserDateOverridesFile`),
    consistent with how `getApiFromConfigFile` is handled.
  - `createdDate` is validated as a valid ISO 8601 date using the same parser (luxon) that the
    date filter relies on, so a non-ISO value fails loudly instead of silently skipping a user.
  - `username` in the file is optional and is not processed — it's only there for readability.

  ## Usage

  Overrides file (`overrides.json`):

  ```json
  {
      "createdDate": "2026-06-10",
      "users": [
          { "id": "uid1" },
          { "id": "uid2", "username": "pepito" }
      ]
  }

  Run:

  yarn start usermonitoring run-2fa-reporter \
    --config-file config.json \
    --disable-users \
    --filtered-by-month \
    --created-date-overrides-file overrides.json
    --created-date-overrides-file has no effect without --filtered-by-month; the script logs a warning if the file is passed without it.

  Test plan

  - [ ] Run without --created-date-overrides-file → existing behaviour unchanged.
  - [ ] Override a user (who would be disabled) with a recent createdDate → user is not disabled.
  - [ ] Override a recently-created user with an old createdDate → user is disabled.
  - [ ] Pass the file without --filtered-by-month → warning is logged, file has no effect.
  - [ ] Pass a malformed / non-JSON file, a file missing createdDate/users, or a non-ISO
  date → script fails with a clear error message.
  - [ ] Unit tests: yarn test-unit (80 tests passing).

### DIFF
--- a/README.md
+++ b/README.md
@@ -617,6 +617,22 @@ An exceptionGroup list containing groups whose users should be excluded from val
 
 An optional --disable-users CLI flag that, when passed, will disable users identified as invalid due to missing or disabled 2FA.
 
+An optional --filtered-by-month (-f) CLI flag that, when passed, restricts disable candidates to users whose account was created at least `disableAfterMonths` months ago (configured in the datastore).
+
+An optional --created-date-overrides-file CLI flag that accepts a path to a JSON file. For each user listed in the file, their real DHIS2 creation date is replaced by the `createdDate` field in the file when evaluating the month-based filter (`--filtered-by-month`). This allows treating specific users as if they were created on a given date. **Only has effect when --filtered-by-month is also set.**
+
+The overrides file format:
+
+```json
+{
+    "createdDate": "2026-06-10",
+    "users": [
+        { "id": "uid1" },
+        { "id": "uid2", "username": "pepito" }
+    ]
+}
+```
+
 The datastore must contain:
 
 ```json
--- a/src/domain/entities/DateTime.ts
+++ b/src/domain/entities/DateTime.ts
@@ -30,3 +30,8 @@ export function getMonthsDiff(startDate: string, endDate: string): number {
     const diff = end.diff(start, "months").months;
     return Math.floor(diff);
 }
+
+/* Return true if the given string is a valid ISO 8601 date/datetime (as parsed by luxon). */
+export function isValidIso8601(date: string): boolean {
+    return DateTime.fromISO(date).isValid;
+}
--- a/src/domain/entities/user-monitoring/two-factor-monitoring/TwoFactorUser.ts
+++ b/src/domain/entities/user-monitoring/two-factor-monitoring/TwoFactorUser.ts
@@ -12,6 +12,11 @@ export interface TwoFactorUser {
     created: Timestamp;
 }
 
+export interface UserDateOverride {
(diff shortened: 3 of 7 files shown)


MatiasArriola (changes requested):
Thanks @idelcano! Overall the PR looks good, and it's nice to see test coverage.

Some code review comments only (I haven't tested): 

- We would need resolving the conflicts with development before merging
- `userMonitorintg.ts` `parseUserDateOverridesFile`: Replace manual json validation with a `Codec`.  For reference see `localConfigCodec` in the same file. Then you can keep the ISO date check as a second semantic validation step, or wrap that in a refined/custom codec.
  -  this would make the code consistent and also validate that the json effectively is correctly shaped (i.e. `users[n].id` is mandatory)
- The PR says `--created-date-overrides-file` has no effect without `--filtered-by-month`, but the file is parsed and validated even without `--filtered-by-month` set.
  - Maybe we can move the overrides parsing behind the `filteredByMonth` guard - not really important but for sake of correctness

idelcano:
> Thanks @idelcano! Overall the PR looks good, and it's nice to see test coverage.
> 
> Some code review comments only (I haven't tested):
> 
> * We would need resolving the conflicts with development before merging
> * `userMonitorintg.ts` `parseUserDateOverridesFile`: Replace manual json validation with a `Codec`.  For reference see `localConfigCodec` in the same file. Then you can keep the ISO date check as a second semantic validation step, or wrap that in a refined/custom codec.
>   
>   * this would make the code consistent and also validate that the json effectively is correctly shaped (i.e. `users[n].id` is mandatory)
> * The PR says `--created-date-overrides-file` has no effect without `--filtered-by-month`, but the file is parsed and validated even without `--filtered-by-month` set.
>   
>   * Maybe we can move the overrides parsing behind the `filteredByMonth` guard - not really important but for sake of correctness

All fixed
Thanks matias!

MatiasArriola (approved):
Thanks @idelcano!

cgbautista (approved):
Works as expected, both to force a recent date and for forcing an old date.

PR #106, opened 2026-07-06
translations to-spreadsheet: --program-id filter and short locale references

tokland:
Closes https://app.clickup.com/t/4528615/869dc4guz?comment=90120236273934

Follow-up to #101, adding two capabilities to `translations to-spreadsheet`.

## Changes

- **`--program-id` filter.** When set, objects are taken from that program's metadata dependency export (`/api/programs/{id}/metadata`) instead of the whole instance, so each generated spreadsheet is scoped to a single program.
- **Short locale references.** `--locales` now matches by exact base name first, then by substring, so `Sotho` resolves to `Southern Sotho (Lesotho)`. A reference matching more than one locale is ambiguous and errors; no match warns and is skipped (as before).
- **Suffix-stripped column headers.** The resolved locale's ` (...)` qualifier is dropped from the column header (`formName: Southern Sotho` instead of `formName: Southern Sotho (Lesotho)`). This also fixes the round-trip: `from-spreadsheet` matches headers against suffix-stripped DB names, so previously any locale with a `(country)` suffix would not re-import.

## Test plan

- Unit tests added/updated in `ExportTranslationsUseCase.spec.ts` (short/substring match, ambiguity error, stripped header, program-scoped fetch). `yarn vitest run` green; `tsc --noEmit` clean.
- Verified end-to-end against a live instance: program-scoped export produces one program's objects; `Sotho` → `Southern Sotho` header; and a `from-spreadsheet --post` round-trip added a `st_LS` translation while preserving existing ones.

### DIFF
--- a/src/data/ExportTranslationsSpreadsheetRepository.ts
+++ b/src/data/ExportTranslationsSpreadsheetRepository.ts
@@ -77,7 +77,9 @@ export class ExportTranslationsSpreadsheetRepository implements ExportTranslatio
 
         _.forEach(files, (content, name) => {
             if (!/^xl\/worksheets\/sheet\d+\.xml$/.test(name)) return;
-            const xml = decoder.decode(content).replace(/<sheetView ([^>]*?)\/>/, `<sheetView $1>${pane}</sheetView>`);
+            const xml = decoder
+                .decode(content)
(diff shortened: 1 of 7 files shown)


tokland:
- [x] Add support for `--data-set-id ID`  (required by https://app.clickup.com/t/4528615/869f24nhh)

PR #108, opened 2026-08-12
[Feature] Add start/end-date, descendants and skip-metadata to export programs script

nshandra:
### :pushpin: References

Issue: [[49] Migrate event data for SKERU](https://app.clickup.com/t/4528615/869edk1gd)

### :memo: Implementation

- Update export programs script to allow the data to be filtered by the `lastUpdated` property using the `--start-date` and/or `--end-date` options.

- Added the  `--descendants` option, is its used, the included orgUnits will be the specified by `--orgunits-ids` and its descendants.

- The metadata export can be skipped via the `--skip-metadata`option.

- Change program import to allow importing without metadata.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ Notes:
 Export a program with all its associated metadata and data (events, enrollments, tracked entities).
 
 The data can be filtered by:
-- Organization Unit via the `--orgunits-ids` option. If `--descendants` is used, the included orgUnits will be the specified and its descendants.
+- Organization Unit via the `--orgunits-ids` option. If `--descendants` is used, the included orgUnits will be the specified and its descendants. Note that if no `--orgunits-ids` is provided the `ouMode` will be `ALL` and the `--descendants` option will be redundant.
 - The `lastUpdated` property using the `--start-date` and/or `--end-date` options.
 
 The metadata export can be skipped via the `--skip-metadata`option.
--- a/README.md
+++ b/README.md
@@ -109,6 +109,13 @@ Notes:
 
 Export a program with all its associated metadata and data (events, enrollments, tracked entities).
 
+The data can be filtered by:
+- Organization Unit via the `--orgunits-ids` option. If `--descendants` is used, the included orgUnits will be the specified and its descendants.
+- The `lastUpdated` property using the `--start-date` and/or `--end-date` options.
+
+The metadata export can be skipped via the `--skip-metadata`option.
+
+
 ```shell
 $ yarn start programs export --url='http://USER:PASSWORD@HOST:PORT' \
   --programs-ids=kX2GpLIa75l,kpNc7KvydVz programs.json
--- a/README.md
+++ b/README.md
@@ -113,7 +113,9 @@ The data can be filtered by:
 - Organization Unit via the `--orgunits-ids` option. If `--descendants` is used, the included orgUnits will be the specified and its descendants. Note that if no `--orgunits-ids` is provided the `ouMode` will be `ALL` and the `--descendants` option will be redundant.
 - The `lastUpdated` property using the `--start-date` and/or `--end-date` options.
 
-The metadata export can be skipped via the `--skip-metadata`option.
+The metadata export can be skipped via the `--skip-metadata` option.
+
+The login details can be either provided as `--url='http://USER:PASSWORD@HOST:PORT'` or `--url='http://HOST:PORT' --auth='USER:PASSWORD'`. The `--auth` option is useful with complex passwords that can cause errors when used in the `--url` only approach.
 
 
 ```shell
--- a/src/data/D2Tracker.ts
+++ b/src/data/D2Tracker.ts
@@ -76,12 +76,14 @@ export class D2Tracker {
             orgUnitIds: string[] | undefined;
             trackedEntity?: string | undefined;
             children?: boolean;
+            startDate?: string;
+            endDate?: string;
         }
     ): Promise<Array<Mapping[Key][number]>> {
         type Output = Array<Mapping[Key][number]>;
 
         const output: Output = [];
-        const { programIds, orgUnitIds, trackedEntity } = options;
+        const { programIds, orgUnitIds, trackedEntity, startDate, endDate } = options;
 
         for (const programId of programIds) {
             let page = 1;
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -109,6 +109,13 @@ Notes:
 
 Export a program with all its associated metadata and data (events, enrollments, tracked entities).
 
+The data can be filtered by:
+- Organization Unit via the `--orgunits-ids` option. If `--descendants` is used, the included orgUnits will be the specified and its descendants. Note that if no `--orgunits-ids` is provided the `ouMode` will be `ALL` and the `--descendants` option will be redundant.
+- The `lastUpdated` property using the `--start-date` and/or `--end-date` options.
+
+The metadata export can be skipped via the `--skip-metadata`option.
```
anagperal: [non-blocking] missing space  `--skip-metadata` option.

src/scripts/commands/programs.ts
```diff
@@ -45,25 +45,47 @@ const programIdsOptions = option({
...
+    long: "end-date",
+    description: "End date",
+});
+
 const exportCmd = command({
     name: "export",
     description: "Export program metadata and data (events, enrollments, TEIs)",
     args: {
-        url: getApiUrlOption(),
+        ...getApiUrlOptions(),
```
anagperal, line 64: Add also this change to README: now `url` + `auth` is also valid instead of all in `url`

src/data/ProgramsD2Repository.ts
```diff
@@ -114,8 +121,10 @@ export class ProgramsD2Repository implements ProgramsRepository {
     }
 
     async import(programExport: D2ProgramExport): Async<void> {
```
anagperal, line 123: `D2ProgramExport.metadata` is still required while the domain `ProgramExport.metadata` is now optional

src/data/D2Tracker.ts
anagperal, line 115: `updatedBefore` isn't supported on the enrollments endpoint, d2-api's `trackerEnrollments.d.ts:59` has `updatedAfter` and `updatedWithin` only. Nothing catches this: assigning to a variable before passing disables TypeScript's excess-property check (`apiOptions`)

anagperal, line 102: FYI I don't know which version of DHIS2 this script is running against, but I'm just leaving a comment here as a heads-up that from version 2.42 onwards, `ouMode` is `orgUnitMode`.

README.md
```diff
@@ -109,6 +109,13 @@ Notes:
 
 Export a program with all its associated metadata and data (events, enrollments, tracked entities).
 
+The data can be filtered by:
+- Organization Unit via the `--orgunits-ids` option. If `--descendants` is used, the included orgUnits will be the specified and its descendants. Note that if no `--orgunits-ids` is provided the `ouMode` will be `ALL` and the `--descendants` option will be redundant.
+- The `lastUpdated` property using the `--start-date` and/or `--end-date` options.
```
anagperal, line 114: Just a comment: the date window is applied independently to three related entity types (events, enrollments and tracked entities), so a filtered export can contain dangling references.

For example, with `--start-date=2026-01-01`: an event edited on 2026-02-14 is exported, but its enrollment (last updated 2024-06-02) and its TEI (2024-05-10) are not, so the file references two records it doesn't contain. If this is intentional like this, then it's ok.

src/scripts/commands/programs.ts
```diff
@@ -45,25 +45,47 @@ const programIdsOptions = option({
     description: "List of program (comma-separated)",
 });
 
+const startDateArg = option({
+    type: optional(string),
+    long: "start-date",
+    description: "Start date",
```
anagperal, line 51: improve description here and in end date

anagperal (changes requested):
[code-only review] thanks @nshandra !! In general it seems ok, just a few comments

README.md
nshandra: Fixed.

src/scripts/commands/programs.ts
nshandra: True, added.

src/data/ProgramsD2Repository.ts
nshandra: Fixed.

src/data/D2Tracker.ts
nshandra: Its for DHIS2.40 for now, but good point. I added a comment.

nshandra: Added a post-query filter in D2Tracker for enrollments that emulates this function.

README.md
nshandra: Added a warning about this in the README. The task didn't warranted tackling this a the data to extract were of event type only.

The fix would be to collect the missing references in the retrieved tracker type metadata and fetch them afterwards.

src/scripts/commands/programs.ts
nshandra: I restored `startDateArg` / `endDateArg` and created new `updatedStartDateArg` / `updatedEndDateArg` with a more detailed description and a input validation function.

src/data/D2Tracker.ts
```diff
@@ -126,6 +131,21 @@ export class D2Tracker {
 
         return output;
     }
+
+    // NOTE: the enrollments endpoint has no updatedBefore param
+    private filterByUpdatedEndDate<Key extends TrackerDataKey>(
```
anagperal: [non-blocking] The post-query filter works. Just a typing detail: the two casts (`instances as apping["enrollments"]` and `as Array<Mapping[Key][number]>`) aren't needed. Events, enrollments and tracked entities all have `updatedAt` in d2-api, so the filter type-checks without them, and an unneeded `as` can hide a future type mismatch from the compiler.
Optionally, the filter could apply to all three models, without the `model` param:

  ```ts
  function filterByUpdatedEndDate<T extends { updatedAt: string }>(
      instances: ReadonlyArray<T>,
      updatedEndDate: string | undefined
  ): T[] {
      if (!updatedEndDate) return [...instances];
      const endDate = new Date(updatedEndDate).getTime();
      return instances.filter(instance => new Date(instance.updatedAt).getTime() < endDate);
  }
  ```

anagperal (approved):
Thanks @nshandra , all the comments are addressed! I'm approving since only a non-blocking comment is left (the unneeded casts in `filterByUpdatedEndDate`), though it would be better to fix it.
Also, please update the PR description: it still mentions `--start-date` / `--end-date`, which no longer exist (now `--updated-start-date` / `--updated-end-date`)

